PR #18, opened 2015-08-01
write a bit smaller SVG

schriftgestalt:
The %g command does not write .0 if there are no decimal places. And
the relative coordinates save a few bytes, too.

### DIFF
--- a/drawBot/context/svgContext.py
+++ b/drawBot/context/svgContext.py
@@ -451,14 +451,18 @@ class SVGContext(BaseContext):
             aT.setTransformStruct_(transformMatrix[:])
             path.transformUsingAffineTransform_(aT)
         svg = ""
+        lastPoint = None
         for i in range(path.elementCount()):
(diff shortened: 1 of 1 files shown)


typemytype:
Thanks Georg, but the %g formatting is not a good option to round off numbers, as it only show the first 4 (in the case) digits of a number.

```Python
print "%.4g" % 1
print "%.4g" % 1.1
print "%.4g" % 1.0
print "%.4g" % 1.23
print "%.4g" % 1.2345
print "%.4g" % 10.2345
print "%.4g" % 100.2345
print "%.4g" % 1000.2345
print "%.4g" % 10000.2345
```
>>>1
1.1
1
1.23
1.234
10.23
100.2
1000
1e+04


But it is indeed important to remove the redundant `.0` digits in the svg file, this is fixed in https://github.com/typemytype/drawbot/commit/a12a725d099fede1a36611baa3f9df9424dbde79

PR #20, opened 2015-08-17
Make the app run on 10.6

karstenw:
py2app v0.10
pyobjc 3.1b1
Python 2.7.9


typemytype:
cool!

does this work with the default installation of py2app and pyOjbc? (the versions which are pre-installed)
Is the specific combination of py2app and pyObjc that makes this work?

what is the difference between `super()` and `objc.super()`? for pure python objects, not for NSObjects.

I don't like to embed fontTools, ufo2svg in the repo. As those have to be updated frequently.

I also would like to keep the context tools near by the context, as DrawBot can also be installed as a module, and I don't want to install those separately. 

thanks

karstenw:
On 17.08.2015, at 10:44, Frederik Berlaen wrote:

> cool!
> 
> does this work with the default installation of py2app and pyOjbc? (the versions which are pre-installed)

I do not use or fiddle with the pre-installed Python.

> Is the specific combination of py2app and pyObjc that makes this work?

I guess so. I was totally surprised this morning by the running of the 10.10 exe on 10.6. This was against all pyobjc-black-magic wisdom I was used to.

In both cases I have compiled Python myself and installed recent versions of pyObjc and py2app. On 10.6 it's strictly 32-bit to avoid complications with external non-universal libs, on 10.10 it's strictly 64-bit since new Frameworks are 64-bit only.


> what is the difference between `super()` and `objc.super()`? for pure python objects, not for NSObjects.

It should work with Python objects but I haven't tested it.

objc.super was a recent change and not so well documented. I found a hint in a forum.

> I don't like to embed fontTools, ufo2svg in the repo. As those have to be updated frequently.

Agreed. I did it just for the convenience as ufo2svg didn't have setup.py and fonttools was handled just the same.


> I also would like to keep the context tools near by the context, as DrawBot can also be installed as a module, and I don't want to install those separately.

This was a hack for gifsicle. I just changed that and also replaced the gifsicle exe with a universal one (not yet tested).

The only real problems involving new code were:
- the color conversion in pdfContext.py
-- _cmykNSColorToCGColor & _rgbNSColorToCGColor both needed an additional color conversion. Otherwise the attributes (red|green|blue|cyan|magenta|yellow|black)Component were not present

- the new functions nsmaxrange(range) nslocationinrange(loc,range) in ui/codeEditor.py
-- There was a mismatch between signed/unsigned 64/32 bit UINT_MAX and the usage of -1


-karsten

karstenw:
fontTools is gone. What about ufo2svg? I included it because it was required, but I've never seen any use of it (as of fontTools).

typemytype:
> I guess so. I was totally surprised this morning by the running of the 10.10 exe on 10.6. This was against all pyobjc-black-magic wisdom I was used to.

That scares me... but if the source is working both for a custom install as for the default, it's fine for me

> objc.super was a recent change and not so well documented. I found a hint in a forum.

link? I didnt found any references, it just looks ugly... and wondering if it is really required to get it working on older OS's.

fontTools is required by roboFab, if you want to work with ufo files inside DrawBot. ufo2svg is one of the convenient packages while working with ufo/fonts. 

(Im testing this out during the next days)

thanks

karstenw:
On 20.08.2015, at 11:23, Frederik Berlaen wrote:

>> I guess so. I was totally surprised this morning by the running of the 10.10 exe on 10.6. This was against all pyobjc-black-magic wisdom I was used to.
> 
> That scares me... but if the source is working both for a custom install as for the default, it's fine for me

I had my custom python 2.7.10 on OSX 10.10.4 configured with:
env MACOSX_DEPLOYMENT_TARGET=10.6 CFLAGS="-arch x86_64 " LDFLAGS="-arch x86_64 " ./configure --enable-framework --enable-ipv6 --enable-toolbox-glue

I think the crucal point is MACOSX_DEPLOYMENT_TARGET=10.6

Another pyobjc (https://github.com/karstenw/Cactus-opml-Outliner) project did not run on 10.6 when built on 10.10 because I lazyly pip-installed an external lib (lxml) that uses libxml which in turn was the 10.10 specific one.

DrawBot has no external (C code) dependencies so presumably this stunt should work.


>> objc.super was a recent change and not so well documented. I found a hint in a forum.
> 
> link? I didnt found any references, it just looks ugly... and wondering if it is really required to get it working on older OS's.

I think it was this one:
http://sourceforge.net/p/pyobjc/mailman/pyobjc-dev/?viewmonth=201405&viewday=30

This is for PyObjc >= 3.0. IIRC DrawBot uses a manually installed version

In older versions the built-in super was silently overridden with the objc.super version.


> fontTools is required by roboFab, if you want to work with ufo files inside DrawBot. ufo2svg is one of the convenient packages while working with ufo/fonts.

I think I'll put ufo2svg manually into site-lib and remove it from the repo.

typemytype:
> _cmykNSColorToCGColor & _rgbNSColorToCGColor both needed an additional color conversion. 

I dont understand how a such a color would end up in a context, they are converted to the specified colorspace on init see https://github.com/typemytype/drawbot/blob/master/drawBot/context/baseContext.py#L280

karstenw:
On 20.08.2015, at 15:59, Frederik Berlaen wrote:

>> _cmykNSColorToCGColor & _rgbNSColorToCGColor both needed an additional color conversion. 
> 
> I dont understand how a such a color would end up in a context, they are converted to the specified colorspace on init see https://github.com/typemytype/drawbot/blob/master/drawBot/context/baseContext.py#L280


The weren't. On 10.6 on several occasions DrawBot would crash here

https://github.com/karstenw/drawbot/blob/master/drawBot/context/pdfContext.py#L324

claiming that NSColor c did not have attribute "cyanComponent".

I went to the docs and they said, the color needs another conversion.

That behavior was never observed on 10.10.4 only on 10.6

I just branched & restored the old functions. Here's the Exception when running "colorSpace.py":

  File "/Users/karstenw/python/github/DrawBot/drawbot/drawBot/ui/drawBotController.py", line 109, in createContext
    _drawBotDrawingTool._drawInContext(context)
  File "/Users/karstenw/python/github/DrawBot/drawbot/drawBot/drawBotDrawingTools.py", line 98, in _drawInContext
    attr(*args, **kwargs)
  File "/Users/karstenw/python/github/DrawBot/drawbot/drawBot/context/baseContext.py", line 1182, in rect
    self.drawPath(path)
  File "/Users/karstenw/python/github/DrawBot/drawbot/drawBot/context/baseContext.py", line 1212, in drawPath
    self._drawPath()
  File "/Users/karstenw/python/github/DrawBot/drawbot/drawBot/context/pdfContext.py", line 113, in _drawPath
    self._pdfFillColor()
  File "/Users/karstenw/python/github/DrawBot/drawbot/drawBot/context/pdfContext.py", line 263, in _pdfFillColor
    cgColor = self._rgbNSColorToCGColor(c)
  File "/Users/karstenw/python/github/DrawBot/drawbot/drawBot/context/pdfContext.py", line 344, in _rgbNSColorToCGColor
    return Quartz.CGColorCreateGenericRGB(c.redComponent(),
ValueError: NSInvalidArgumentException - *** -redComponent not defined for the NSColor NSCustomColorSpace sRGB IEC61966-2.1 colorspace 0.791591 0.606354 1 1; need to first convert colorspace.

README.md
```diff
@@ -1,4 +1,10 @@
 DrawBot
 =======
 
-DrawBot is a powerful, free application for MacOSX that invites you to write simple Python scripts to generate two-dimensional graphics. The builtin graphics primitives support rectangles, ovals, (bezier) paths, polygons, text objects and transparency.
\ No newline at end of file
+DrawBot is a powerful, free application for MacOSX that invites you to write simple Python scripts to generate two-dimensional graphics. The builtin graphics primitives support rectangles, ovals, (bezier) paths, polygons, text objects and transparency.
+
+
+This is a private fork just for the purpose of making a 10.6 executable. [Download from my Dropbox.](https://goo.gl/22OHVh)
```
adrientetar: You might want to drop this from the PR:

karstenw: Why would I want that?

adrientetar: The part that says "this is a private fork," as making a PR against mainline is not private fork.

typemytype:
fyi: I didnt had time yet to look into this as it will require some time to install all the specified dependencies, but it will happen

justvanrossum:
I'm afraid this pull request is no longer viable :(

karstenw:
> I'm afraid this pull request is no longer viable :(

That was a misnomer back then. It should have been named "How to make DrawBot self containing so it could be made to run on 10.6"

I hear self containing apps are currently a hot topic...

PR #33, opened 2016-01-25
Building on El Capitan

jenskutilek:
I tried to build DrawBot on El Cap, and had to change the "super" function to use objc.super for classes inheriting from Cocoa classes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ docs/_build
 local
 
 *.command
+env/*
--- a/DrawBot.py
+++ b/DrawBot.py
@@ -12,6 +12,7 @@ from drawBot.misc import getDefault, stringToInt
 from drawBot.updater import Updater
 
 import objc
+from objc import super
 
 objc.setVerbose(True)
 
--- a/drawBot/ui/codeEditor.py
+++ b/drawBot/ui/codeEditor.py
@@ -1,5 +1,6 @@
 import AppKit
 import objc
+from objc import super
 
 from keyword import kwlist
 import re
--- a/drawBot/ui/lineNumberRulerView.py
+++ b/drawBot/ui/lineNumberRulerView.py
@@ -1,5 +1,6 @@
 from AppKit import *
 import math
+from objc import super
 
 
 """
--- a/drawBot/ui/splitView.py
+++ b/drawBot/ui/splitView.py
@@ -1,4 +1,5 @@
 from AppKit import *
+from objc import super
 
 from vanilla import *
 from vanilla.vanillaBase import VanillaBaseObject


typemytype:
this is a similar request as @karstenw did, see https://github.com/typemytype/drawbot/pull/20
would this also make DrawBot run on 10.6+, could you test?

thanks

jenskutilek:
I will try to find a 10.6 machine and test ;)

karstenw:
I have a 10.6 machine. Just add an url to a binary.

jenskutilek:
Here's a binary: https://www.dropbox.com/s/gfpf75i2lh3evzv/drawbot.zip?dl=0

Let me know if it works on 10.6.

jenskutilek:
On El Capitan, the System Integrity Protection (SIP) interferes with the py2app build, because the SIP "protected" status is copied with the files while building the app bundle. Then you suddenly don't have write access to your own app bundle, and thus the build process aborts.

I worked around this by setting up a virtual Python environment and building with that. I hope this doesn't complicate things further ;)

karstenw:
It does not run on 10.6.

The first obstacle is the Info.plist "Minimum system version" with a value of 10.9. Tweaked to 10.6. Please delete this entry. It is not needed.

You should get rid off the aliased version. The python executable is soft linked to your virtenv. That way it'll only run on your machine.

```
% ll ./DrawBot.app/Contents/MacOS/
total 216
drwxr-xr-x@ 4 karstenw  staff     136 Jan 28 16:50 .
drwxr-xr-x@ 7 karstenw  staff     238 Jan 28 17:02 ..
-rwxr-xr-x@ 1 karstenw  staff  103952 Jan 28 16:36 DrawBot
lrwxr-xr-x@ 1 karstenw  staff      48 Jan 28 16:50 python -> /Users/jenskutilek/Python/drawbot/env/bin/python
```

On 10.10. I saw some problems too when building an app. I resolved it by building several times untill the last line said the build was ok.

You could try to compile my Drawbot fork [https://github.com/karstenw/drawbot](https://github.com/karstenw/drawbot) It compiles on 10.6 and 10.10 but misses the changes since 2015-08.

jenskutilek:
I can't find a way to include a full Python environment in the app bundle. Followed a [tutorial](https://www.metachris.com/2015/11/create-standalone-mac-os-x-applications-with-python-and-py2app/) and the resulting app still only contains a python binary linked to the virtual environment. This seems to have changed between versions (of Mac OS X or py2app?), as the newest comment in the tutorial hints at.

Without a virtual environment, the app cannot be built due to SIP ... this is where I'm out of ideas ...

karstenw:
After some googling around:

### sip disabling is needed for el capitan & py2app
https://forums.developer.apple.com/thread/6987

### how to disable sip
http://osxdaily.com/2015/10/05/disable-rootless-system-integrity-protection-mac-os-x/

When I read that stuff I'm glad that 10.6 is still my workhorse and 10.10 is so annoying that I boot it only once a month to build some py2apps.

justvanrossum:
@jenskutilek,, @typemytype, how relevant is this PR today? Should we close it?

jenskutilek:
@justvanrossum what system version are you building DrawBot on? I guess if we can build on 10.12 and 10.13, we may close this. But I expect the modification of the `super` function to be needed there as well.

In the meantime I found out that the SIP problem goes away if you install an additional copy of `py2app`, ignoring the pre-installed module:

```bash
$ pip install --user --ignore-installed py2app
```

typemytype:
you can build on any system...
from 10.11+ its advised to install a custom python with all required packages
the system python is will only work on that machine.. (these are my finding after given up the system python after a day...)

jenskutilek:
I successfully built from the master branch on 10.12, so I guess we can close this PR. I got a warning in the console though, perhaps it is important:

```Detected missing constraints for <private>.  It cannot be placed because there are not enough constraints to fully define the size and origin. Add the missing constraints, or set translatesAutoresizingMaskIntoConstraints=YES and constraints will be generated for you. If this view is laid out manually on macOS 10.12 and later, you may choose to not call [super layout] from your override. Set a breakpoint on DETECTED_MISSING_CONSTRAINTS to debug. This error will only be logged once.```

PR #36, opened 2016-01-28
Fix stupid mistake (but it probably was never triggered)

jenskutilek:
Sorry, I made a stupid mistake in the last pull request.


typemytype:
did it manually see https://github.com/typemytype/drawbot/commit/efded845112ee1311a000adffdca430bc2a42e0f

Issue #47, opened 2016-03-30
AVFoundation


typemytype:
if someone is in the right mood to convert the QTKit mov movie context to AVFoundation
see https://github.com/typemytype/drawbot/blob/master/drawBot/context/movContext.py

https://developer.apple.com/library/mac/technotes/tn2300/_index.html

justvanrossum:
This would still be nice, but in the meantime we have `.mp4` export which is better and faster than the current `.mov` export.

justvanrossum:
Here's a Swift example: https://gist.github.com/JaleelNazir/ce17ac24ec99636a230b3de8a98d7000

justvanrossum:
But it seems to require CoreVideo, which is not included in PyObjC I think (can't find it).

justvanrossum:
Oh, except that all those things are accessible via Quartz.

typemytype:
closing in favor of the current ffmpeg implementation

PR #48, opened 2016-05-19
More accurate movie timing by using a ratio for the frameDuration

andyclymer:
More accurate frame timing is possible when using a ratio for the frameDuration, rather than a floating point, since the QTMovie ends up needing a ratio for the frame duration. I made it so that the old API works too, the frameDuration can either be given as a float or as a tuple.

fps = 60
frameDuration(1 / fps) # as a float
frameDuration((1, fps)) # as a ratio

I found that the rounding error when converting from a float back into a ratio can result in timing problems in the movie when trying to sync the movie to events in an audio track.

The attached Python script will aim to make two 30-second long 60fps movies. Open them up in QuickTime Player and check the Inspector window: the movie made with a floating point frameDuration ends up as 27-seconds long and 66.67fps, but the movie with the frameDuration as a ratio ends up with the correct length and rate.

Thanks!!
Andy

[frameDurationTest.py.zip](https://github.com/typemytype/drawbot/files/271863/frameDurationTest.py.zip)

### DIFF
--- a/drawBot/context/gifContext.py
+++ b/drawBot/context/gifContext.py
@@ -28,6 +28,9 @@ class GifContext(ImageContext):
         self._delayData = []
 
     def _frameDuration(self, seconds):
+        if type(seconds) is tuple:
+            # Duration as a ratio, turn it into a fraction of a second
+            seconds = seconds[0] / seconds[1]
         # gifsicle -h: Set frame delay to TIME (in 1/100sec).
         self._delayData[-1] = int(seconds * 100)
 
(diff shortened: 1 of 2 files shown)


justvanrossum:
I will look into that. The floating point cannot be the issue directly, so I wonder whether we can fix the precision without changing the API.

justvanrossum:
I've committed a different fix that works well for most common frame rates: https://github.com/typemytype/drawbot/commit/66437db488933384c3abd1ab1011138e307bb916
What I can't get to work is a framerate of 29.97, even if I hardcode QTKit.QTMakeTime(100, 2997), so that's a little odd. But given the deprecated status of QTKit we probably should waste too much time on it.

typemytype:
A while ago Ive made a note to switch to AVFoundation see https://github.com/typemytype/drawbot/issues/47

andyclymer:
Thanks you guys!

PR #53, opened 2016-07-03
Added the ability to create links within a PDF

andyclymer:
Add a link rect with the name of a link on one page, and a link destination location with the same name on another page. The (x,y) location of the link destination defaults to None, in which case the link destination will be placed in the center of the page.

### DIFF
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -1361,6 +1361,13 @@ class BaseContext(object):
 
     def _printImage(self, pdf=None):
         pass
+    
+    def _linkDestination(self, name, (x, y)):
+        pass
+
+    def _linkRect(self, name, (x, y, w, h)):
+        pass
+    
 
     ###
 
@@ -1763,3 +1770,9 @@ class BaseContext(object):
         if psName is not None:
             psName = psName.toUnicode()
         return psName
+
+    def linkDestination(self, name, (x, y)):
+        self._linkDestination(name, (x, y))
+    
+    def linkRect(self, name, (x, y, w, h)):
+        self._linkRect(name, (x, y, w, h))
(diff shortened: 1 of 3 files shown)


typemytype:
Should a formatted string also be able to support urls/links?

andyclymer:
That's probably a good idea. Should I take a shot at it? It looks like in the pdfContext there's a moment when I'd be able to get the bounding box for a text box as it's being drawn? I'd have to hunt around a bit but I could try :)

typemytype:
I would like to add `linkText(string, (x, y), urlStr)`

This should also be added to a formattedString. 
This could be stored in the NSAttributedString and used while drawing. The rect of a text with a link could be retrieved with [CTRunGetImageBounds or CTRunGetTypographicBounds](https://developer.apple.com/library/ios/documentation/Carbon/Reference/CTRunRef/index.html#//apple_ref/c/func/CTRunGetImageBounds)

if links are going to be supported `underline`, `strikethrough` and maybe some other text decorations options should be supported....

Links could be really cool: fe to write interactive specimens, and each glyph has a link with robofont-script://<pathToScript.py> so glyphs could be edited directly in RoboFont with a click from a pdf...

cc @justvanrossum

justvanrossum:
Re. `linkDestination()` and `linkRect()`: the documentation is rather sparse. From reading the doc strings I have no idea what this does. What is the `name` argument? How to use these functions?

justvanrossum:
Also, the word `link` suggests URL, which it isn't.

typemytype:
it could be a `name` or an `URL`

A name will jump from the `linkRect` to the `linkDestination`

When an URL is provided it will open the it in the default browser. (see [CGPDFContextSetURLForRect](https://developer.apple.com/library/mac/documentation/GraphicsImaging/Reference/CGPDFContext/#//apple_ref/c/func/CGPDFContextSetURLForRect))

Testing the difference between a `name` and and `URL` is maybe as simple as testing on 
`"://" in nameOrLink`

PR #81, opened 2017-01-11
Add alignment to text().

yanone:
Previously, text() would use whatever alignments have been used for
textBox() without being able to explicitly revert/set the alignment for
text(). Only an empty textBox() with an alignment would set the alignment
for text().

### DIFF
--- a/drawBot/drawBotDrawingTools.py
+++ b/drawBot/drawBotDrawingTools.py
@@ -1011,7 +1011,7 @@ class DrawBotDrawingTool(object):
 
     # drawing text
 
-    def text(self, txt, x, y=None):
+    def text(self, txt, x, y=None, align=None):
         """
         Draw a text at a provided position.
 
@@ -1037,7 +1037,7 @@ class DrawBotDrawingTool(object):
         if origins:
             x -= origins[-1][0]
             y -= origins[-1][1]
-        self.textBox(txt, (x, y-h, w*2, h*2))
+        self.textBox(txt, (x, y-h, w*2, h*2), align=align)
 
     def textOverflow(self, txt, box, align=None):
         """


justvanrossum:
The previous behavior is definitely a bug. But with the patch applied things are not fixed completely.

Commenting out the `textBox()` call changes the behavior of `text()`:

```python
fontSize(50)

textBox("a bb ccc " * 20, (100, 100, 600, 600), align="right")

rect(500, 0, 10, 40)
text("Hey", (500, 50), align="left")

```

Also, when `align='right'` in `text()` it doesn't align the way I would expect.

yanone:
It right-aligns at the (x, y) point (at least over here). I would expect this. Since text() doesn’t carry a size but only an origin point, the text starts at that point, going left-wards.
That is also exactly the same behaviour as in Illustrator, for instance, when comparing single-line text and text boxes.

typemytype:
this is not aligned correctly:
screen shot 2017-01-11 at 18 09 45

I understand the need for an `align` in `text(...)`

typemytype:
solved here: https://github.com/typemytype/drawbot/commit/fc84562f76c4eb32f2ad847684c235aa8dd5b6ac

please test

yanone:
Using that produces left-aligned text for me.
And looking at the code it seems that `w` is used 8 lines before it’s set. But I receive no traceback for that. It goes through, only left-aligned.

yanone:
Sorry, forget my previous reply. It does throw a traceback for `w` and it does right-align after I’ve commented those lines out.
So just the `w` needs to be fixed.

typemytype:
oh fixed here https://github.com/typemytype/drawbot/commit/54822afac46a5d648b1a3d4de1ba72ee3e1f52a7

yanone:
Yup, now it works. Thank you.

PR #89, opened 2017-01-19
moving examples to docstrings

gferreira:
¡Warning! A few issues still need to be fixed:

1. with `::downloadcode`, special characters like `\n` and `\t` get converted into the actual newline and tab characters – this broke some examples in which those characters appear, like `indent.py` and `tabsFormattedString.py`

2. the ‘downloads’ folder is included in the git repo – this is redundant since the examples are now in the docstrings. (easy to fix by adding the ‘downloads’ folder to `.gitignore`)

3. the ‘downloads’ folder should probably be cleared when running `make html`, or the new download files should simply overwrite existing ones – otherwise you end up with `rect.py`, `rect1.py`, `rect2.py` etc.


typemytype:
sphinx escapes "\n" with "\\n"

maybe not so nice for `help(fun)` but the download and docs works as it should

gferreira:
are you sure the docs are ok? it looks like the sample code is breaking at `\n` here:

https://github.com/typemytype/drawbot/pull/89/commits/76e44c917c6371fabe66777469ef96e48947b4e0#diff-b2fc94f9cc832e46572e3e9d70505ed1

screen shot 2017-01-19 at 21 19 12

typemytype:
double \\\n slashes

github was also escaping them :)

docs/downloads/indent.py
```diff
@@ -38,29 +38,4 @@
 # add text
 t += txt
 # add hard return
-t += "\n"
```
typemytype: this should be `\\n` to get properly escaped

gferreira:
thanks – fixed! (issues 2 & 3 remain though)

gferreira:
https://github.com/typemytype/drawbot/pull/89/commits/20d2a663b7bd2ed9a0d9a03ba7041048202d7ec9 addresses point 2 (see above)

typemytype:
super thanks!!!

PR #104, opened 2017-05-18
Consolidate three README files into one.

frankrolf:
I consolidated the files so they all show on the main GitHub landing page (rather than having to click through different readme files).
Hope it makes sense.
I tried to distinguish the three sectors as clearly as possible.


typemytype:
thanks, that makes idd sense to have only 1 read me...

PR #111, opened 2017-08-09
Fixing FormattedString appendGlyph() when using a font from a path

andyclymer:
FormattedString appendGlyph() needs to _tryInstallFontFromFontName() just like append() does, otherwise it won't use a font installed from a path

### DIFF
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -1430,10 +1430,14 @@ class FormattedString(object):
         baseString = unichr(0x00A0)
         font = None
         if self._font:
-            font = AppKit.NSFont.fontWithName_size_(self._font, self._fontSize)
-        if font is None:
-            warnings.warn("font: %s is not installed, back to the fallback font: %s" % (self._font, _FALLBACKFONT))
-            font = AppKit.NSFont.fontWithName_size_(_FALLBACKFONT, self._fontSize)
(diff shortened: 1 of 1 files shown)


typemytype:
thanks Andy

but that is already happening, when setting a font it tries to install and retreive and store the font name in `formattedString._font`

see https://github.com/typemytype/drawbot/blob/master/drawBot/context/baseContext.py#L1079-L1081

if this is incorrect, please reopen or change the pull request

andyclymer:
Hi Frederik! It looks like there's still a problem, I think this sample code will help show what's going on:

```
import drawBot as db

db.newDrawing()
db.size(100, 100)

# This works, I can appendGlyph when the font is called by name
txt1 = db.FormattedString()
txt1.append("abc", font="Helvetica", fontSize=20, fill=(1, 0, 0))
txt1.appendGlyph("dollar") # This works fine
txt1.append("123")
db.text(txt1, (10, 50))

# This doesn't work, the appended glyph doesn't show up when I specify the font from its path
txt2 = db.FormattedString()
txt2.append("def", font="/path/to/font.otf", fontSize=20, fill=(0, 0, 1))
txt2.appendGlyph("dollar") # This glyph won't be drawn
txt2.append("456")
db.text(txt2, (10, 20))

db.saveImage("test.pdf")
```

typemytype:
lets continue the discussion in that issue

PR #125, opened 2017-11-08
Various py3 changes; should all still work on py2.7, too

justvanrossum:
*** This has only been lightly tested. ***

It's a simple port: you build DrawBot either with Python 2.7, and then you code DB with 2.7, or you build it with Python 3.6.*, and then you code DB with that version of Python.

Things to look into:
- how does importing external module work, if at all?
- I didn't touch the auto updater, so that may cause some trouble

It would be nice to offer it as an experimental download at drawbot.com, named DrawBotPy3.app.

### DIFF
--- a/drawBot/drawBotDrawingTools.py
+++ b/drawBot/drawBotDrawingTools.py
@@ -79,7 +79,7 @@ class DrawBotDrawingTool(object):
 
     def _get_version(self):
         try:
-            import drawBotSettings
+            from drawBot import drawBotSettings
             return drawBotSettings.__version__
         except Exception:
             pass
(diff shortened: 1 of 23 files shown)


justvanrossum:
Btw. one tiny py3 tweak to vanilla was needed: https://github.com/typesupply/vanilla/commit/d1b8e42d09681e97f9ff16cc98804afbf3435f8a

drawBot/drawBotDrawingTools.py
```diff
@@ -58,7 +62,7 @@ def _deprecatedWarningWrapInTuple(txt):
     '10x14'       : (720, 1008),
 }
 
-for key, (w, h) in _paperSizes.items():
+for key, (w, h) in list(_paperSizes.items()):
```
adrientetar: list can be removed safely here

justvanrossum: Well I put it there because I got an exception. I'll double check.

justvanrossum: Yeah, because the loop adds items to the dict, so without the `list()` there will be an `RuntimeError: dictionary changed size during iteration` exception.

adrientetar: oh, my bad you're right!

anthrotype:
Good work! :tada:

PR #129, opened 2017-11-25
add mp4 export support via ffmpeg

justvanrossum:


typemytype:
but pffff 35mb :)

justvanrossum:
Ja, sorry... But still far less than the difference between the .dmg for Py2 and Py3 :)

typemytype:
woop woop!

justvanrossum:
Oh: all command line tools end up twice in the app binary, once in Resources, and once zipped with with drawBot package :( So there is a lot of waste currently.

justvanrossum:
The ones from drawBot.context.tools are also included, but useless as they are inside a zip file.

karstenw:
On 25.11.2017, at 22:36, Just van Rossum wrote:

> The ones from drawBot.context.tools are also included, but useless as they are inside a zip file.

The external tools belong in the "data_files" section of the setup() command.

Like this:

dataFiles = [
    "Resources/English.lproj",
    "drawBot/context/tools/gifsicle",
    ...

...
setup(
    data_files=dataFiles,
    ....

I have the impression that the setup.py makes a lot of unnecessary copying.


-karsten

typemytype:
this is true, I will make a separate issue for this
see https://github.com/typemytype/drawbot/issues/130

typemytype:
maybe what is missing is to read a .mp4 and get a specific frame, similar to an animated .gif

see https://github.com/typemytype/drawbot/blob/master/drawBot/context/tools/gifTools.py#L93

I dont know if this is actually a thing, I dont want to overcomplicate things...

justvanrossum:
Does that work for .mov? I didn't think so :)

Joking aside, it's probably possible via ffmpeg, but I personally don't need it, and I don't think we're ready for a more general reading-of-video-material API.

typemytype:
no mov reading so no mp4 reading, thats easy ;)

mov support maybe has to go completly... as it is slow and using deprecated cocoa apis

PR #137, opened 2017-12-11
Updating print statements to be Python 3 compatible

meyouwe:
Updating courseware.rst and quickReference.rst so that the print statements are Python 3 compatible.

### DIFF
--- a/docs/content/courseware.rst
+++ b/docs/content/courseware.rst
@@ -22,36 +22,36 @@ How do numbers look in Python? how to write simple sums, multiplication, substra
 
     # this is a comment
 
-    print "some basic numbers:"
-    print 12  # this is an integer number
-    print 12.5  # this is a floating point
+    print("some basic numbers:")
+    print(12)  # this is an integer number
+    print(12.5)  # this is a floating point
 
-    print "results of adding:"
-    print 12 + 13  # results in an integer
-    print 12 + 0.5  # results in a float
-    print 0.5 + 12  # ditto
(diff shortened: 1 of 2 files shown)


justvanrossum:
Thanks, Maarten! Good to see you here :)

It gets a little ugly when there's more than one argument to `print()`: without `__future__` statement these now behave differently in Python 2.7. On the one hand, I'd like to avoid throwing `__future__` directives at beginners, on the other hand, I don't really see a way around it,,,

typemytype:
thanks Maarten!!

@justvanrossum: I would rather leave `__future__` out of the entry level py introduction, `print(...)` just works in the py2 version of the app. Maybe it should just split up multiple argument printing into multiple print statements?

justvanrossum:
Yes, if that doesn't obfuscate the examples too much, that's good, too.

meyouwe:
Hi all. On New Zealand time here. Sorry I didn't realise this would be an issue. I tested the changes I made in the two docs in Drawbot 3.107 and had no errors so thought this would be fine. The reason I changed them was because when I cut and paste from the docs on the drawbot.com site I got an error. I have updated my Drawbot to 3.110 and noticed that `print '...'` and `print(...)` both work fine.

justvanrossum:
I'll merge tomorrow and will touch up the multi-item prints, where needed.

justvanrossum:
I want to wait until #153 is merged, and then I'll merge this.

PR #138, opened 2017-12-12
Adding context-specific options to the saveImage() function

justvanrossum:
This PR (which is work in progress) adds the possibility to pass context-specific options via the saveImage() function.

Documentation needs to be done!

Right now there are the following options:
- `imageResolution`: specifies which resolution to render when creating pixel output (default is still at 72)

Some properties from NSImage are mapped, see https://developer.apple.com/documentation/appkit/nsbitmapimagereppropertykey?language=objc
- `imageColorSyncProfileData`
- `imageJPEGCompressionFactor`
- `imageTIFFCompressionMethod`
- `imageGIFDitherTransparency`
- `imageFallbackBackgroundColor`
- `imagePNGGamma`
- `imagePNGInterlaced`
- `imageJPEGProgressive`
- `imageGIFRGBColorTable`

And there's (so far) one ffmpeg-specific option:
- `ffmpegCodec`: specify which codec to use when making mp4 with ffmpeg


drawBot/context/imageContext.py
```diff
@@ -35,13 +49,18 @@ def _writeDataToFile(self, data, path, multipage):
             firstPage = pageCount - 1
             pathAdd = ""
         outputPaths = []
+        imageResolution = options.get("imageResolution", 72.0)
+        properties = {}
+        for dbKey, nsKey in _nsImageOptions.items():
+            if dbKey in options:
+                properties[nsKey] = options[dbKey]
```
typemytype: there are some converters needed: if a nscolor is required it would be nice just to provide `(1, 0, 0)`

drawBot/drawBotDrawingTools.py
```diff
@@ -363,7 +363,7 @@ def pages(self):
                     break
         return tuple(DrawBotPage(instructionSet) for instructionSet in instructions)
 
-    def saveImage(self, paths, multipage=None):
+    def saveImage(self, paths, multipage=None, **options):
```
typemytype: would it be an idea to move `multipage` to options too?
I understand this is done to keep scripts with `saveImage(path, True)` working
but multipage feels like option :) where a pdf and image context are interested in...

drawBot/context/imageContext.py
justvanrossum: Yes, but that seems to be the only one. I'll fix that.

drawBot/drawBotDrawingTools.py
justvanrossum: Yeah, let's not break the api, but it is indeed a context-specific option. I'll see how to unify this.

Together with docs, it would be nice if we could validate the option names. Right now any misspelling is simply ignored.

typemytype: `*args` could just capturing `multipage` and warn and put it inside the options 
`def saveImage(self, paths, *deprecatedArgs, **options)`

related to validation and docs: each context could have a `ctx.saveImageOptions` a list of all possible options (similar as `fileExtensions`)

drawBot/context/imageContext.py
typemytype: and maybe the NSData keys, drawBot would like to provide raw data, not an NSData object, actually I dont know what kind of data should be provided.

drawBot/context/imageContext.py
```diff
@@ -8,17 +8,28 @@
 from .pdfContext import PDFContext
 
 
+def _nsDataConverter(value):
+    if isinstance(value, AppKit.NSData):
+        return value
+    return AppKit.NSData.dataWithBytes_length_(value, len(value))
+
+def _nsColorConverter(color):
```
typemytype: or reuse a color object from the base context https://github.com/typemytype/drawbot/blob/master/drawBot/context/baseContext.py#L619

this will support directly all possible color tuple combinations `(0, )`,  `(1, .5)`,  `(1, 0, 0)`, `(1, 0, 0, 1)`

justvanrossum: I can do that. I'm confused about `NSColor.colorWithDeviceRed_green_blue_alpha_()` vs. `NSColor.colorWithCalibratedRed_green_blue_alpha_()`, though. the `Color` object uses the latter, but the former matches `imagePixelColor()` more precisely.

drawBot/drawBotDrawingTools.py
justvanrossum: `ctx.saveImageOptions` now exists, and is a list of `(optionName, optionDoc)` tuples. Validation is done in the public `saveImage()` entry point. @typemytype can you see how this can be integrated into the docs?

justvanrossum: I left the public `saveImage()` function signature as it is (only adding `**options`). I prefer that over adding a deprecation warning. I'm fine with people using `multipage` as a positional argument, let's just not do it in our examples.

drawBot/drawBotDrawingTools.py
```diff
@@ -406,8 +406,14 @@ def saveImage(self, paths, multipage=None):
             context = getContextForFileExt(ext)
             if context is None:
                 raise DrawBotError("Could not find a supported context for: '%s'" % ext)
+            allowedSaveImageOptions = set(optionName for optionName, optionDoc in context.saveImageOptions)
```
typemytype: this has to be tested outside the loop as this is possible:
multiple files and each with their own options... 
`saveImage([path.png, path.pdf, path.mp4], imagePNGGamma=5, ffmpegCodec="mpeg4")`

drawBot/drawBotDrawingTools.py
```diff
@@ -405,10 +409,6 @@ def saveImage(self, paths, multipage=None, **options):
             context = getContextForFileExt(ext)
             if context is None:
                 raise DrawBotError("Could not find a supported context for: '%s'" % ext)
-            allowedSaveImageOptions = set(optionName for optionName, optionDoc in context.saveImageOptions)
-            for optionName in options:
-                if optionName not in allowedSaveImageOptions:
-                    warnings.warn("Unrecognized saveImage() option found for %s: %s" % (context.__class__.__name__, optionName))
```
justvanrossum: A) I don't care much for the path = list option. I never use it. I find it much easier to use separate saveImage() calls
B) I did it inside the loop so that the warning message is specific for the context being used. To specify `ffmpegCodec` when saving `.png` should cause a warning, but not when saving `.mp4`. Please revert.

drawBot/context/baseContext.py
```diff
@@ -1677,7 +1677,9 @@ class BaseContext(object):
     _gradientClass = Gradient
 
     fileExtensions = []
-    saveImageOptions = []
+    saveImageOptions = [
+        ("multipage", None),
+    ]
```
justvanrossum: I left that list empty because I only want `multipage` to be documented for context that actually do something with it. And why would the doc string be None?

tests/testExport.py
```diff
@@ -152,19 +152,19 @@ def _testMultipage(self, extension, expectedMultipageCount):
...
-        self._testMultipage(".svg", 5)
+        self._testMultipage(".svg", numFrames=5, expectedMultipageCount=5)
 
     def test_multipage_gif(self):
-        self._testMultipage(".gif", 0)
+        self._testMultipage(".gif", numFrames=5, expectedMultipageCount=0)
 
     def test_multipage_pdf(self):
-        self._testMultipage(".pdf", 0)
+        self._testMultipage(".pdf", numFrames=5, expectedMultipageCount=0)
```
justvanrossum: What is being fixed in this commit?

typemytype: it only worked when the amount of frame was set to 5, any other number of frames would assert

drawBot/drawBotDrawingTools.py
typemytype: A) I understand :)

B) fe: when setting a different `imagePNGGamma` for generated png's to build a  mp4 would show me a warning?

I guess either solution are optimal as saveImage is maybe combining to much :)

justvanrossum: Good point, but perhaps the mp4 context should then list those options as well.

tests/testExport.py
justvanrossum: Did the tests fail???

drawBot/context/imageContext.py
```diff
@@ -32,19 +32,25 @@ def _tiffCompressionConverter(value):
...
+    "imageFallbackBackgroundColor": (AppKit.NSImageFallbackBackgroundColor, _nsColorConverter,         "The background color to use when writing to an image format (such as JPEG) that doesn't support alpha. The color's alpha value is ignored. The default background color, when this property is not specified, is white. The value of the property should be an NSColor object or a DrawBot RGB color tuple."),
+    "imageGIFDitherTransparency":   (AppKit.NSImageDitherTransparency,      None,                      "Boolean that indicates whether the image is dithered"),
+    "imageGIFRGBColorTable":        (AppKit.NSImageRGBColorTable,           _nsDataConverter,          "A bytes or NSData object containing the RGB color table."),
+}
+
+
+def getSaveImageOptions(options):
+    return ImageContext.saveImageOptions + [
+        (dbKey, _nsImageOptions[dbKey][-1]) for dbKey in options if dbKey in _nsImageOptions
+    ]
```
justvanrossum: `options` is unused, do you mean `options` instead of `ImageContext.saveImageOptions`?

drawBot/drawBotDrawingTools.py
```diff
@@ -443,7 +443,7 @@ def saveImage(self, path, *args, **options):
         context = getContextForFileExt(ext)
         if context is None:
             raise DrawBotError("Could not find a supported context for: '%s'" % ext)
-        if not context.lazySaveImageOptions:
+        if not context.validateSaveImageOptions:
```
justvanrossum: The `not` needs to go...

Issue #141, opened 2017-12-15
Implement command-period cancelling while script is running


justvanrossum:
I've looked at running a thread that would peek into the event queue, and would then post a SIGINT signal to the main thread, but I can't get that to work: I can't seem to access the main thread events from a secondary thread.

What will work is this: use a trace function, to be used in `sys.settrace(mytrace)`. However, that causes quite some overhead (things slow down by a factor of 1.5 - 2). But if we implement such mechanism in C, the overhead should be _much_ smaller, as at the C level the trace function is just a C function, no Python overhead at all. Will need to experiment with that.

justvanrossum:
Hm, even a C trace function slows down the running code by about a factor of two. A pure Python no-op function actually slows down the code by a factor of 3.

typemytype:
could this be a preference?

justvanrossum:
Absolutely.

justvanrossum:
Ok, I'm back to the old solution: run a thread which polls the event queue. The problem with that is twofold:
- Carbon's `CheckEventQueueForUserCancel()` has no Cocoa equivalent. (`NSApp().nextEventMatchingMask_...` doesn't work from a thread)
- Python doesn't support Carbon anymore, so `Evt.CheckEventQueueForUserCancel()` no longer exists

However, Carbon.framework is still there, and with ctypes I can actually use it after all.

Whipping up a PR now that pretty much reimplements the old DrawBot solution.

PR #142, opened 2017-12-15
While a script is running, run a thread to monitor command-period

justvanrossum:
...and interrupt the main thread if the user indeed wants to cancel the execution of the script.

This pulls in a function from Carbon.framework via ctypes. I'm not sure how future-proof that is. But so far, this works like a champ.

### DIFF
--- a/drawBot/scriptTools.py
+++ b/drawBot/scriptTools.py
@@ -21,7 +21,11 @@ from fontTools.misc.py23 import PY2, PY3
 from drawBot.misc import getDefault
 
 # Pulling in CheckEventQueueForUserCancel from Carbon.framework
-CheckEventQueueForUserCancel = ctypes.CFUNCTYPE(ctypes.c_bool)(('CheckEventQueueForUserCancel', ctypes.CDLL(find_library("Carbon"))))
+_carbonPath = find_library("Carbon")
+if _carbonPath is not None:
+    CheckEventQueueForUserCancel = ctypes.CFUNCTYPE(ctypes.c_bool)(('CheckEventQueueForUserCancel', ctypes.CDLL(_carbonPath)))
+else:
+    CheckEventQueueForUserCancel = None
 
 # Acquire this lock if something must not be interrupted by command-period or escape
 cancelLock = threading.Lock()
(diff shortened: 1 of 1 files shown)


justvanrossum:
This implements #141.

justvanrossum:
Btw. I can't measure any slowdown because of the monitoring thread.

typemytype:
I will make a test build so this can be tested on other OS's...
This workaround to get `CheckEventQueueForUserCancel ` from Carbon library feels a bit old fashioned. 

I know there is nothing to replace this call. Executing the script in a separate thread is also not a good options, as all UI related things would not work if they are not executed the main thread.

justvanrossum:
Executing the script in a thread also doesn't work for another reason: you can't interrupt a secondary thread with a SIGINT signal, or at all really.

Another alternative is to run the script in a separate process, and pipe the resulting PDF data back to the app.

So there are three ways we can make a script interruptable:
- use Carbon's `CheckEventQueueForUserCancel()` from a thread, post SIGINT to main thread. This doesn't seem to have much of a performance penalty. Some memory usage fro the monitor thread.
- use `sys.trace()` / `PyEval_SetTrace()`. The performance hit in the script is significant, a factor of about 2 was the lowest I managed.
- run the script in a separate process. That is safest and most efficient (will run in a separate core even), but allows no interaction with direct UI calls. (The `Variable()` UI would be tricky, but I see ways of doing it.)

justvanrossum:
If we go with the Carbon function, I will make the code robust for the absence of `Carbon.framework,` and just continue without the cancelling capability.

justvanrossum:
I just read that `Carbon/framework` is still there on 10.13.

justvanrossum:
`CheckEventQueueForUserCancel()` is indeed a bit of a stop-gap measure, but it's so much better than nothing. I've missed command-period ever since DB 3, and I lost count of how many unsaved changes I saw getting lost because force-quit was the least unreasonable way out...

typemytype:
+100 for the better than nothing approach

justvanrossum:
So let's merge? :)

Issue #144, opened 2017-12-16
Embedded image() doesn't work in SVG output


justvanrossum:

typemytype:
see https://github.com/typemytype/drawbot/blob/master/drawBot/context/svgContext.py#L456-L458

typemytype:
that could be a `saveImage` setting for a svg context: `embedImages`

maybe the same for `embedFonts`??

justvanrossum:
Right now it embeds an url to an absolute local path, so no matter how you look at it the current behavior is broken:
```svg
 <image height="660.0" opacity="1" transform="matrix(0.378787878788,0.0,0.0,0.378787878788,0.0,250.0)" width="660.0" x="0" xlink:href="/Users/just/code/git/drawbot/tests/data/drawBot.jpg" y="0"/>
```

typemytype:
is there an other option then embedding? saving the image relative to the svg file?

justvanrossum:
That would probably work (that's what the href is for I think), but I doubt that's what our users would expect. I think embedding by default is fine, possibly augmented by disabling it and using relative paths with a saveImage() option.

justvanrossum:
There may be limitations anyway. Without looking it up, I don't expect embedding PDF data is supported.

justvanrossum:
Maybe for now we should just warn `*** embedding images in SVG is currently not supported ***`...

justvanrossum:
Btw. it's not just that the referenced image can't be found. This is the error from Safari:
```
This page contains the following errors:

error on line 3 at column 54: Namespace prefix xlink for href on image is not defined
Below is a rendering of the page up to the first error.
```
Firefox says something similar.

typemytype:
a pdf could be converted to a bitmap and be embedded

typemytype:
this is merged!
well done!!

PR #145, opened 2017-12-16
Remove deprecated lowercase api

justvanrossum:
...as well as deprecated argument styles, such as `line(1, 2, 3, 4)` instead of `line((1, 2), (3, 4))`.


justvanrossum:
(That cwd commit in retrospect doesn't belong here, but that was before I discovered #144. I meant to add `image()` tests here, too.)

PR #149, opened 2017-12-18
Save context manager

typemytype:


drawBot/drawBotDrawingTools.py
```diff
@@ -465,6 +465,16 @@ def save(self):
...
+        class SaveContextManager(object):
+
+            def __enter__(saveCxt):
+                return saveCxt
+
+            def __exit__(saveCxt, type, value, traceback):
+                self.restore()
+
+        return SaveContextManager()
+
```
justvanrossum: I would write it like so:
```python
# at global level
class SaveContextManager(object):
    def __init__(self, drawingTools):
        self._drawingTools = drawingTools
    def __enter__(self):
        return self
    def __exit__(self type, value, traceback):
        self._drawingTools.restore()

# in save():
    return SaveContextManager(self)
```

typemytype: I didnt want to keep a reference of a callback or an object somewhere...
this feels very similar as a decorator wrapper

justvanrossum: Why not? To not have `self` as a first argument is very ugly. It's an explicit reference as opposed to a scoped variable, which I think is rather obscure when using a nested class.

justvanrossum:
Trying to document this and writing an doc string example makes me doubt the name again. Let's sleep on it, perhaps a different name is easier to explain. Would `savedState()` be an option?

```python
rect(0, 0, 100, 100)
with savedState():
    translate(100, 100)
    rotate(45)
    rect(0, 0, 100, 100)
rect(100, 100, 100, 100)
```

justvanrossum:
A different name makes it easier to refer to good old save/restore and vice versa. Eventually save/restore may be deprecated, but that will be harder if we overload save() with a context manager.

justvanrossum:
I slept over it and am now strongly in favor of `savedState()`. I can implement that later, unless you beat me to it.

drawBot/drawBotDrawingTools.py
```diff
@@ -496,6 +496,26 @@ def restore(self):
...
+        """
+        Save and restore the current state in a `with` statement.
+
+        .. downloadcode:: savedState.py
+
+            # use with statement
+            # this will be wrapped around a `save` and `restore`
+            with savedState():
+                # set a color
+                fill(1, 0, 0)
```
justvanrossum: I would love a transformation in the example, too, like rotate and/or translate. I hardly ever use (or need) a saved graphics state for paint styles, but all the time for transformations.

justvanrossum:
```
======================================================================
ERROR: test_rect (testExamples.ExampleTester)
----------------------------------------------------------------------
  File "/Users/just/code/git/drawbot/tests/testExamples.py", line 143, in test
    self.assertFilesEqual(imagePath, expectedImagePath)
  File "/Users/just/code/git/drawbot/tests/testExamples.py", line 69, in assertFilesEqual
    self.assertEqual(readData(path1), readData(path2), "Files %r and %s are not the same" % (path1, path2))
  File "/Users/just/code/git/drawbot/tests/testSupport.py", line 119, in readData
    with open(path, "rb") as f:
IOError: [Errno 2] No such file or directory: '/Users/just/code/git/drawbot/tests/data/example_rect.png'

======================================================================
FAIL: test_lineDash (testExamples.ExampleTester)
----------------------------------------------------------------------
  File "/Users/just/code/git/drawbot/tests/testExamples.py", line 143, in test
    self.assertFilesEqual(imagePath, expectedImagePath)
  File "/Users/just/code/git/drawbot/tests/testExamples.py", line 69, in assertFilesEqual
    self.assertEqual(readData(path1), readData(path2), "Files %r and %s are not the same" % (path1, path2))
AssertionError: Files '/Users/just/code/git/drawbot/tests/tempTestData/example_lineDash.png' and /Users/just/code/git/drawbot/tests/data/example_lineDash.png are not the same
```

typemytype:
does this one needs more editing?

perhaps test could be added later in an other PR

PR #150, opened 2017-12-18
SVG: replace uuid id generating with deterministic method

justvanrossum:
The use of uuid is problematic when testing: I didn't find a seed method for those random ids. I replaced it with a deterministic counter approach per id type.


justvanrossum:
Also, this adds tests for all gradient types.

typemytype:
maybe add also such a unique id gen for images

justvanrossum:
Embedded images don't need it as the context keeps track of them.
`imageID = "image_%s" % (len(self._embeddedImages) + 1)`

Issue #127, opened 2017-11-13
DrawBotPy3: the entire Python std library needs to be embedded.


justvanrossum:
As a little todo for Python3 (where we don't have a local Python available).

http://learning-python.com/pyedit-products/unzipped/build/build-app-exe/macosx/include-full-stdlib.py
http://learning-python.com/pyedit-products/unzipped/build/build-app-exe/macosx/setup.py

typemytype:
this list https://github.com/jackmaney/python-stdlib-list/blob/master/stdlib_list/lists/3.6.txt
can be inserted in  https://github.com/typemytype/drawbot/blob/master/setup.py#L104
only for the PY3 version...

and I guess the list can also be smaller: I dont feel to need to support `email`, `mailbox`, `telnetLib`, `tkinter`.... and some more, maybe we have to look at the size?

justvanrossum:
That list seems to be made with the code I reference above, and is far from complete, it's missing a lot of `_underscore` modules. Yest, `tkinter` doesn't make sense, because it won't work together with Cocoa, but for the rest I really want to offer a complete Python stdlib. Why wouldn't someone want to program some email visualisations with DrawBot? I don't want to make any assumptions.

I'm working on it, will push a PR today.

justvanrossum:
Implemented via https://github.com/typemytype/drawbot/commit/876359559a05916bf128856915107656ba19eb08

PR #156, opened 2017-12-22
For non-/System builds of DB, include the entire Standard Lib

### LINKED ISSUE #148: colorsys module is missing in py3 build
Hey Just! Hey Frederik!

I just noticed that colorsys module is not included into the Drawbot py3 build.
Was this intended?

That module is quite helpful for converting colors between RGB to HLS and back.
Could you please consider to include it again?

Cheers

justvanrossum:
Implements #127 and fixes #148.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,11 +2,13 @@ from __future__ import print_function
 
 import py2app
 from distutils.core import setup
+from distutils.sysconfig import get_python_lib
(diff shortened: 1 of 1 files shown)


justvanrossum:
FWIW, this added about 11MB to the dmg. Could be way worse :)

Issue #123, opened 2017-11-06
method intersection()


RafalBuchner:
I think that BezierPath's method called intersection is buggy: I couldn't use it and I don't know how can I repair it.

this is output from my DrawBot:

Traceback (most recent call last):
  File "intersection_problem.py", line 13, in <module>
  File "drawBot/context/baseContext.pyc", line 499, in intersection
  File "/Applications/DrawBot.app/Contents/Resources/lib/python2.7/booleanOperations/booleanOperationManager.py", line 112, in intersection
  File "/Applications/DrawBot.app/Contents/Resources/lib/python2.7/booleanOperations/booleanOperationManager.py", line 64, in _performOperation
  File "/Applications/DrawBot.app/Contents/Resources/lib/python2.7/booleanOperations/flatten.py", line 64, in __init__
  File "drawBot/context/baseContext.pyc", line 52, in drawToPointPen
  File "/Applications/DrawBot.app/Contents/Resources/lib/python2.7/booleanOperations/flatten.py", line 342, in addPoint
AssertionError

here is my code:
>    stroke(0)
>    fill(None)
>    bezier = BezierPath()
>    bezier.moveTo((100,100))
>    bezier.curveTo((100,200),(200,200),(200,100))
>    drawPath(bezier)

>    bezier2 = BezierPath()
>    bezier2.moveTo((222,200))
>    bezier2.curveTo((200,100),(10,200),(50,100))
>    drawPath(bezier2)

>    bezier2.intersection(bezier)

justvanrossum:
Can't check right now, but it could be because your paths are note closed. (If so, the error message could be better, yes...)

justvanrossum:
Also, intersection() will return a new path if I'm not mistaken.

typemytype:
What is your DrawBot version? I guess this is an older one...

boolean operations on paths requires to have closed paths

this works fine in the current version

```python
stroke(0)
fill(None)
bezier = BezierPath()
bezier.moveTo((100,100))
bezier.curveTo((100,200),(200,200),(200,100))
bezier.closePath()
drawPath(bezier)

bezier2 = BezierPath()
bezier2.moveTo((222,200))
bezier2.curveTo((200,100),(10,200),(50,100))
bezier2.closePath()

drawPath(bezier2)

result = bezier2.intersection(bezier)

fill(1, 0, 0, .5)
stroke(None)
drawPath(result)
```

[...]

PR #159, opened 2017-12-22
add listIntersections

typemytype:
+ test

see #123


justvanrossum:
Maybe simly call it `bez.intersections()`?

PR #166, opened 2018-01-04
don't use apples prefix

schriftgestalt:


justvanrossum:
Thanks, I noticed that, too, and didn't get round to fix that.

justvanrossum:
All sane non-Cocoa frameworks that subclass NSObject use a prefix that is not NS or UI...

schriftgestalt:
The conversion in Cocoa is to use a private (mostly) two letter prefix (because there is no namespace in objectiveC). So the correct Cocoa way would be to use the prefix for all NSObject subclasses. Having the super class in the class name is a bit unusual.

justvanrossum:
I'll change it, though.

PR #167, opened 2018-01-04
don’t do `from AppKit import *`

schriftgestalt:
I hope I got everything. Please run tests, I can’t build DrawBot.app.


justvanrossum:
Thanks Georg! I'll test.

justvanrossum:
Works just fine, thanks again! DB on Py 2.7 starts up as fast as the Py3 version. And it seems the Py3 version also got a little quicker to start up.

typemytype:
super("!")

schriftgestalt:
Now someone needs to edit the code samples to not include the slow line.

PR #174, opened 2018-01-19
[WIP] messing with colors, trying to fix #171

### LINKED ISSUE #171: RGB Color in exported images not identical to fill() args
Somehow RGB colors get calibrated along the way. I've tried to play with different ways of constructing the NSColor in baseContext.py, but I don't notice any change. I tried to set colorSpace() to various values, to no effect, either. Is there some way we can gain exact control over the RGB values in exported images?

Oh, I see I haven't played with the color space name in `imagePixelColor()`. Will do that later, and compare values from PIL.

```python
canvasSize = 400
newPage(canvasSize, canvasSize)

bands = 4
bandWidth = canvasSize / bands

fontSize(10)

for i in range(bands):
    x = i * bandWidth
    for j in range(bands):
        y = j * bandWidth
        r = i / bands
        g = j / bands
        b = 0.5
        fill(r, g, b)
        rect(x, y, bandWidth, bandWidth)
        fill(0)
        text("%s,%s,%s" % (r, g, b), (x + 3, y + 5))

fn = "color_blocks1.png"
saveImage(fn)

# 'genericRGB', 'adobeRGB1998', 'sRGB', 'genericGray', 'genericGamma22Gray'

# colorSpace("sRGB")

for i in range(bands):
    x = (i + 0.5) * bandWidth
    for j in range(bands):
        y = (j + 0.5) * bandWidth
        pix = imagePixelColor(fn, (x, y))
        print(pix)
```

justvanrossum:
This PR indeed fixes #171, but unintentionally breaks the `colorSpace()` function. See the doco example of `colorSpace()`.


justvanrossum:
This seems to be the wrong kind of fix.

PR #175, opened 2018-01-19
Another attempt at fixing color issue #171

### LINKED ISSUE #171: RGB Color in exported images not identical to fill() args

justvanrossum:
This seems to fix #171, but currently has a mysterious test failure on Py2, with an `ImageObject()`. Maybe the color space should be an optional argument to `_makeBitmapImageRep()`.


justvanrossum:
I tried to use the device color space in ImageObject, but it makes things worse.

justvanrossum:
I added a test case for `imagePixelColor()`, and that concludes this PR I think.

I had to change `StdOutCollector` along the way, because the list subclass approach doesn't work across py2 and py3, when doing `print(1, 2, 3)`.

justvanrossum:
This PR currently has some minor conflicts with the master branch (caused by me in #177 :) ), I will resolve them once the PR is approved by @typemytype.

justvanrossum:
(Nevermind, conflict has been resolved.)

PR #180, opened 2018-01-22
improve getting the current state of the current font open type feature settings

typemytype:


justvanrossum:
- the tests fail because the `drawBotScripts/openTypeFeatures.py` expected text output has not been updated (beware of dict order diff between py2 and py3)
- traceback when an unknown font is used:
```
font("NotAFont")
print(listOpenTypeFeatures())

  File "<untitled>", line 2, in <module>
  File "/Users/just/code/git/drawbot/drawBot/drawBotDrawingTools.py", line 1514, in listOpenTypeFeatures
    return self._dummyContext._state.text.listOpenTypeFeatures(fontName)
  File "/Users/just/code/git/drawbot/drawBot/context/baseContext.py", line 1301, in listOpenTypeFeatures
    currentFeatures.update(self._openTypeFeatures)
AttributeError: 'list' object has no attribute 'update'
```

- `listFeatureVariations()` should get the same treatment

justvanrossum:
I'm playing with some Devanagari fonts and it seems CoreText doesn't list the OT features that are needed for language support.

For example:

```python
font("ITFDevanagari-Medium")
print(openTypeFeatures())
```
prints an empty dict, but the fonts in `/Library/Fonts/ITFDevanagari.ttc` do contain many GSUB and GPOS features.

Maybe this is for the better, because it may mean we don't have worry about feature defaults that depend on language settings or (worse) text heuristics.

In the context of DB we are mostly interested in (un)setting optional typographic features anyway, so I think it's not a big deal.

typemytype:
it is possible those features are not [listed in this list](https://github.com/typemytype/drawbot/blob/master/drawBot/context/tools/openType.py#L377) (this is donated/taken from harfbuzz)

typemytype:
but coreText is also not listing them...

```python
import CoreText

fontName = "ITFDevanagari-Medium"

descriptor = CoreText.NSFontDescriptor.fontDescriptorWithName_size_(fontName, 12)
featureDescriptions = CoreText.CTFontDescriptorCopyAttribute(descriptor, CoreText.kCTFontFeaturesAttribute)

print(featureDescriptions)
```

justvanrossum:
The tests in `tests/drawBotScripts/fontVariations.py` and `tests/drawBotScripts/openTypeFeatures.py` still fail.

typemytype:
will update the test

maybe a we have to figure out how to travis auto test the tests...

justvanrossum:
While Travis support would be nice, running `$ python3 tests/runAllTests.py` ain't that hard :)

justvanrossum:
- why does `listFontVariations()` return the state? (the `value` key)
- the tests still fail on py2, due to dict order differences, better do something like `print(sorted(d.items()))`.

justvanrossum:
Also, `fontVariations()` should not return min/max/default. This should work and be a no-op:

`fontVariations(**fontVariations())`

Just like `openTypeFeatures(**openTypeFeatures())`

typemytype:
mmmm, I understand the idea:

```python
# that should be possible to round trip (similar to openTypeFeatures(..) and listOpenTypeFeatures()
var = listFontVariations()
fontVariations(var)
```

but variations needs min max (and maybe default) just to know and calculate actual values...

justvanrossum:
No, that is not the idea...

1. `listOpenTypeFeatures()` and `listFontVariations()` give you all the options and default values that are available for the current font. 
2. with `openTypeFeatures()` and `fontVariations()` you set/modify the current settings, and they return the current settings, as a dict that can be applied to themselves as keyword arguments.

To me, this means that `listFontVariations()` returns a dict like:
```
{'wght': {'name': 'Weight', 'minValue': 0.4799, 'maxValue': 3.1999, 'defaultValue': 1.0},
 'wdth': {'name': 'Width', 'minValue': 0.6199, 'maxValue': 1.2999, 'defaultValue': 1.0}}
```
And that `fontVariations()` returns a dict like:

```
{'wght': 1.0, 'wdth': 1.0}
```

typemytype:
so no current state values for `listFontVariations()` `listOpenTypeFeatures()` this feels also weird and unexpected 

and for `fontVariations()` I really would like to know what that value means in a context of a specific axis so you need min max...

typemytype:
openTypeFeatures are simple flags, on off 
fontVariations are a bit more complex and need more context

the return value can differently

justvanrossum:
Ohh it's really a bit of a mess now. I assumed `listOpenTypeFeatures()` returned only default flags, but I see that it returns the current state as well. That makes it impossible to discover what the default state is. I don't like this at all.

So I stick to my idea:
- `listFontVariations()` and `listOpenTypeFeatures()` should return the _possibilities_ and the _default_ state. It's about discovering what's possible with the currently selected font, and what the state would be by default.
- 'fontVariations()` and `openTypeFeatures()` are about the _current_ state. You change the state with them, and they return the current state.

`listFontVariations()` needs to return min/max/default, because it's about discovery. `fontVariations()` can return just the current values, there's no need for it to return min/max/default. Why would there be more context needed there?

The return values of `fontVariations()` and `openTypeFeatures()` should be symmetrical with their (keyword) arguments (apart from the reset flag).

justvanrossum:
The way you propose those functions leaves very little difference between the return values of `listOpenTypeFeatures()` and `openTypeFeatures()` (same for their variations counterparts). I think that's causing redundancy while adding confusion.

typemytype:
they are exactly the same... :)

will change this and adopt the docs
(I did not thought about the option of having `fontVariations()` and `openTypeFeatures()` without arguments... but that makes sense

justvanrossum:
Not sure how far you are with changes, but the following:

```python
font("Helvetica")
print(listOpenTypeFeatures())
font("HoeflerText-Regular")
print(listOpenTypeFeatures())
openTypeFeatures(lnum=True)
print(listOpenTypeFeatures())
font("Helvetica")
print(listOpenTypeFeatures())
```

results in:

```
{'liga': True, 'tnum': True, 'pnum': False}
{'liga': True, 'dlig': False, 'tnum': True, 'pnum': False, 'titl': True, 'onum': True, 'lnum': False}
{'liga': True, 'dlig': False, 'tnum': True, 'pnum': False, 'titl': True, 'onum': True, 'lnum': True}
{'liga': True, 'tnum': True, 'pnum': False, 'lnum': True}
```

This shows (in the third output line) that `listOpenTypeFeatures()` still returns state, and even carries that state on if the font changes to a font that doesn't support that feature (the fourth output line). 

That's another argument for `listXYZ()` not returning state: I really only want to see what the current font supports, not what maybe was set for a different font earlier.

justvanrossum:
Oops sorry, I thought I had pulled in your changes, but something went wrong. I'll try again!

justvanrossum:
Sorry again, it seems everything works according to plan! Excellent.

justvanrossum:
(Except the tests still fail on Python 2.)

justvanrossum:
Still getting this traceback in some cases:

```
  File "/Users/just/code/git/drawbot/drawBot/drawBotDrawingTools.py", line 71, in __enter__
    self._drawingTools.save()
  File "/Users/just/code/git/drawbot/drawBot/drawBotDrawingTools.py", line 484, in save
    self._dummyContext.save()
  File "/Users/just/code/git/drawbot/drawBot/context/baseContext.py", line 1883, in save
    self._stack.append(self._state.copy())
  File "/Users/just/code/git/drawbot/drawBot/context/baseContext.py", line 1701, in copy
    new.text = self.text.copy()
  File "/Users/just/code/git/drawbot/drawBot/context/baseContext.py", line 1481, in copy
    new = self.__class__(**attributes)
  File "/Users/just/code/git/drawbot/drawBot/context/baseContext.py", line 849, in __init__
    self._setAttribute(key, value)
  File "/Users/just/code/git/drawbot/drawBot/context/baseContext.py", line 857, in _setAttribute
    method(**value)
  File "/Users/just/code/git/drawbot/drawBot/context/baseContext.py", line 1291, in openTypeFeatures
    currentFeatures.update(self._openTypeFeatures)
AttributeError: 'list' object has no attribute 'update'
```

typemytype:
this does not show up in the tests...

justvanrossum:
It does now :)

PR #185, opened 2018-02-09
Adds a memoize decorator to cache frequently used getter functions

andyclymer:
I noticed that the last commit in "master" slowed down some larger proofs of mine, I think the problem is just that it needs to repeatedly call some helper functions in openType.py every time you append to a FormattedString.

Would you consider a fix like you find in this commit, that uses a memoize decorator to cache the results? It's one that I had found somewhere in the past and have been using in my own code.

I think this is safe to cache the results from these "get" functions since the font isn't going to change while the script is running.

Example script:

```python
import drawBot as db
import time

fs = db.FormattedString()
runs = 100
totalTime = 0
for i in range(runs):
    now = time.time()
    fs.append("Test", font="Operator-Medium", openTypeFeatures=dict(smcp=True, liga=True))
    totalTime += (time.time() - now)
print totalTime/runs
```
With this script, the timer reports:
``0.0019614815712`` seconds with commit d924098 (before OT feature changes)
``0.0038596010208`` seconds with fd27983 (latest master)
``0.0001614809036`` with caching in openType.py and variations.py

The difference might seem trivial, but a very very large proof now goes from taking 154 seconds to draw, down to 44 seconds. Nice!

### DIFF
--- a/drawBot/context/tools/openType.py
+++ b/drawBot/context/tools/openType.py
@@ -1,4 +1,5 @@
 import CoreText
+from drawBot.misc import memoize
 
 # https://developer.apple.com/fonts/TrueType-Reference-Manual/RM09/AppendixF.html
 # https://github.com/behdad/harfbuzz/blob/master/src/hb-coretext.cc#L381
@@ -470,12 +471,14 @@ for key, value in _featureMap.items():
     reversedFeatureMap[(featureType, featureSelector)] = key
 
 
+@memoize
 def getFeatureTagForFontAttribute(attribute):
     featureType = attribute.get(CoreText.NSFontFeatureTypeIdentifierKey)
     featureSelector = attribute.get(CoreText.NSFontFeatureSelectorIdentifierKey)
     return reversedFeatureMap.get((featureType, featureSelector))
 
 
+@memoize
 def getFeatureTagsForFontAttributes(attributes):
     featureTags = list()
     for attribute in attributes:
@@ -485,6 +488,7 @@ def getFeatureTagsForFontAttributes(attributes):
     return featureTags
 
 
+@memoize
 def getFeatureTagsForDescriptions(featureDescriptions):
     featureTags = dict()
     for featureDescription in featureDescriptions:
@@ -498,6 +502,7 @@ def getFeatureTagsForDescriptions(featureDescriptions):
     return featureTags
 
 
+@memoize
 def getFeatureTagsForFontName(fontName):
     descriptor = CoreText.NSFontDescriptor.fontDescriptorWithName_size_(fontName, 12)
     featureDescriptions = CoreText.CTFontDescriptorCopyAttribute(descriptor, CoreText.kCTFontFeaturesAttribute)
--- a/drawBot/context/tools/variation.py
+++ b/drawBot/context/tools/variation.py
@@ -1,5 +1,6 @@
 import CoreText
 from collections import OrderedDict
+from drawBot.misc import memoize
 
 """
 https://developer.apple.com/documentation/coretext/ctfont/font_variation_axis_dictionary_keys?language=objc
@@ -7,6 +8,7 @@ https://developer.apple.com/documentation/coretext/1508650-ctfontdescriptorcreat
 """
 
 
+@memoize
 def convertIntToVariationTag(value):
     chars = []
     for shift in range(4):
@@ -14,6 +16,7 @@ def convertIntToVariationTag(value):
     return "".join(reversed(chars))
 
 
+@memoize
 def convertVariationTagToInt(tag):
     assert len(tag) == 4
     i = 0
@@ -23,6 +26,7 @@ def convertVariationTagToInt(tag):
     return i
 
 
+@memoize
 def getVariationAxesForFontName(fontName):
     axes = OrderedDict()
     font = CoreText.CTFontCreateWithName(fontName, 12, None)
--- a/drawBot/misc.py
+++ b/drawBot/misc.py
@@ -325,6 +325,30 @@ def getExternalToolPath(root, toolName):
     return toolPath
 
 
+def memoize(function):
+    """
+    Memoize a function's return value with the function's arguments.
+    The next time a function is called with the same arguments, the cache is returned. 
+
+    Example usage:
+        @memoize
+        def addNumbers(first, second):
+            return first + second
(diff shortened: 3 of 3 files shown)


drawBot/context/tools/variation.py
```diff
@@ -1,19 +1,22 @@
...
 from collections import OrderedDict
+from drawBot.misc import memoize
 
 """
 https://developer.apple.com/documentation/coretext/ctfont/font_variation_axis_dictionary_keys?language=objc
 https://developer.apple.com/documentation/coretext/1508650-ctfontdescriptorcreatecopywithva?language=objc
 """
 
 
+@memoize
```
justvanrossum: I bet this one and the next (`convertVariationTagToInt`) make no significant contribution to the speedup.

drawBot/context/tools/openType.py
```diff
@@ -470,12 +471,14 @@
...
 
+@memoize
 def getFeatureTagForFontAttribute(attribute):
     featureType = attribute.get(CoreText.NSFontFeatureTypeIdentifierKey)
     featureSelector = attribute.get(CoreText.NSFontFeatureSelectorIdentifierKey)
     return reversedFeatureMap.get((featureType, featureSelector))
 
 
+@memoize
 def getFeatureTagsForFontAttributes(attributes):
```
justvanrossum: This one (and the one above) are only used for svg export it seems, so should not influence your timing test at all.

drawBot/context/tools/openType.py
```diff
@@ -485,6 +488,7 @@ def getFeatureTagsForFontAttributes(attributes):
     return featureTags
 
 
+@memoize
 def getFeatureTagsForDescriptions(featureDescriptions):
```
justvanrossum: This one only gets called by `getFeatureTagsForFontName()`, which is already being memoized. So memoizing this one doesn't make sense.

drawBot/context/tools/openType.py
justvanrossum: Also, I bet you didn't test this, because memoize arguments need to be hashable, which is not the case here, so this can't work.

drawBot/context/tools/openType.py
justvanrossum: Also, the `featureDescriptions` object comes from CoreText, should be a CFArrayRef, and I doubt it's hashable in a meaningful way (it may compare and hash by identity, which makes it unsuitable as a cache key).

andyclymer:
Thanks for the advice @justvanrossum ! It was a quick hack that definitely sped things back up, but I'll look a little bit deeper into a better way to do it.

typemytype:
... but this shows a really issue: the time to append a short string is became slower.
and only to test if a given font has a feature. If the font does not support a feature it shows a warning. see https://github.com/typemytype/drawbot/blob/master/drawBot/context/baseContext.py#L961

the callback `getFeatureTagsForFontName` could be cached inside a formattedString. The cache could be cleared when a font is installed or reinstalled
see https://github.com/typemytype/drawbot/blob/master/drawBot/context/tools/openType.py#L501

Im thinking about options...

typemytype:
this could work
 
`clearMemoizeCache()` could be called whenever a reset of a drawingTool is called.

this could work out of the box for `getFeatureTagsForFontName` as fontName is a simple string.

```python

_memoizeCache = dict()

def clearMemoizeCache():    
    _memoizeCache.clear()


def memoize(function):
    """
    Memoize a function's return value with the function's arguments.
    The next time a function is called with the same arguments, the cache is returned. 
    Example usage:
        @memoize
        def addNumbers(first, second):
            return first + second
        # The first time this function is called the calculation will be made,
        # and and the result will be stored in the cache dict as [first, second]: returnValue
        # From then on, this value will be returned when the same argument is made to the addNumbers function
    """
    def wrapper(*args):
        if function not in _memoizeCache:
            _memoizeCache[function] = dict()            
        saved = _memoizeCache[function]
        isHashable = True
        for arg in args:
            try:
                hash(arg)
            except TypeError:
                isHashable = False
                break     
        if isHashable and args in saved:
            print("cached")
            return saved[args]
        else:
            print("---- called ----")
            rv = function(*args)
            if isHashable:
                saved[args] = rv
            return rv
    return wrapper
    

@memoize
def myFunction(test, o=None):
    return test * 3


print(myFunction(10, dict()))
print(myFunction(10, dict()))
print(myFunction(10))
print(myFunction(10))
print(myFunction("a"))
print(myFunction("a"))
clearMemoizeCache()
print(myFunction(10))
print(myFunction(10))
print(myFunction(10))
```

typemytype:
or it could be simpler where we dont test if args are hashable... and we know for sure that in case of the `getFeatureTagsForFontName` the args consists out of hashable objects, like a string

justvanrossum:
If you memoize a function, it's fine to assume hashability of the arguments. No need for a check.

Careful: the original memoize cache is obviously per function, and can't be a global dict, so resetting the cache will be more complex than your example code.

typemytype:
maybe we should just call it `ContextMemoize` so its 'clear' it will be all be cleared when a new context is set...

typemytype:
a more optimized version:

```python
_memoizeCache = dict()

def clearMemoizeCache():    
    _memoizeCache.clear()


def memoize(function):
    """
    Memoize a function's return value with the function's arguments.
    The next time a function is called with the same arguments, the cache is returned. 
    Example usage:
        @memoize
        def addNumbers(first, second):
            return first + second
        # The first time this function is called the calculation will be made,
        # and and the result will be stored in the cache dict as [first, second]: returnValue
        # From then on, this value will be returned when the same argument is made to the addNumbers function
    """
    def wrapper(*args):
        key = (function, args)
        if key in _memoizeCache:
            return _memoizeCache[key]
        else:
            result = function(*args)            
            _memoizeCache[key] = result
            return result
    return wrapper
    

@memoize
def myFunction(test, o=None):
    return test * 3


print(myFunction(10, ))
print(myFunction(10, (10, 20)))
print(myFunction(10))
print(myFunction(10))
print(myFunction("a"))
print(myFunction("a"))
clearMemoizeCache()
print(myFunction(10))
print(myFunction(10))
print(myFunction(10))
```

typemytype:
closing in favor of https://github.com/typemytype/drawbot/commit/216e604ca58bbd5c8be11cc3572f7dcba49394ed

thanks!!

PR #186, opened 2018-02-14
improve code editor

typemytype:


drawBot/ui/codeEditor.py
```diff
@@ -589,6 +590,32 @@ def keyDown_(self, event):
...
+            if char in autoCloseMap:
+                selectedText = self.string().substringWithRange_(selectedRange)
+                selectedText = selectedText.strip()
+
+                try:
+                    nextChar = self.string().substringWithRange_((selectedRange.location, 1))
+                    nextChar = nextChar.strip()
+                    if nextChar in autoCloseMap.values():
+                        nextChar = ""
+                except Exception:
```
justvanrossum: What exactly are you catching here?

typemytype: when the string is smaller then range with the same location and a length of 1

line 602-604 and could be outside the try except

typemytype:
on delete char: check if the char is in the autoCloseMap and the nextChar is in the closeChar of the deleted char ---> delete both char

drawBot/ui/codeEditor.py
justvanrossum: So there's two issues here:
- the exception caught is too broad (`Exception` is at the base of the exception hierarchy, so it's pretty much a catch all, potentially masking bugs)
- there are irrelevant lines within the `try`

drawBot/ui/codeEditor.py
```diff
@@ -683,6 +710,19 @@ def insertNewline_(self, sender):
...
 
     def deleteBackward_(self, sender):
+        languageData = self.languagesIDEBehaviorForLanguage_(self.lexer().name)
+        if languageData:
+            selectedRange = self.selectedRange()
+            if selectedRange.length == 0:
+                try:
+                    char = self.string().substringWithRange_((selectedRange.location-1, 1))
+                    nextChar = self.string().substringWithRange_((selectedRange.location, 1))
+                except Exception:
```
justvanrossum: Same thing here. It's a really bad practice and we should avoid it at all cost.

drawBot/ui/codeEditor.py
justvanrossum: Consider `try: ... except TheExactExceptionToBeCaught: ... else: ...`

justvanrossum:
On behavior I like better in TextMate (don't know about other editors) is that if you _do_ "manually" type a closing parenthesis that is already there, it doesn't add another parenthesis, but just moves the cursor to after the closing parenthesis. In short, this means that if you type the actual character sequence (including the chars that are automatically closed) you end up with exactly that char sequence, and the autoclosing business is not getting in your way.

Right now, if I type these characters:

`foo(a)`

the result is

`foo(a))`, with the cursor between the two closing parens.

In TextMate, the result would be

`foo(a)` with the cursor after the closing paren.

justvanrossum:
Another useful subtlety I notice in TextMate: when writing a comment or a string literal, typing a single quote doesn't autoclose it, so you can type apostrophes that are not doubled.

drawBot/ui/codeEditor.py
typemytype: it seems to be an `IndexError`

typemytype:
those last to features are more complex, it has to analyse the line and make a decision. 

* for the `foo(a)` to `foo(a))` it has to search for a open ( sequence ... pff
* for # ' it has the check if the text is been added inside a comment ... also pff

maybe a new those need a new issue for later reference...

justvanrossum:
I prefer to leave this PR open instead of merging it half-finished. It's not an easy feature to get right and saying "pff" doesn't make that go away :) I mean, I really prefer to not have the feature at all over having it misbehave (== get in my way) in obvious cases. Why rush this anyway?

justvanrossum:
Not sure it relates to this PR, but when you type an arrow key in the "Debugger" window, a traceback appears.

```
  File "/Users/just/code/git/drawbot/drawBot/ui/codeEditor.py", line 593, in keyDown_
    languageData = self.languagesIDEBehaviorForLanguage_(self.lexer().name)
AttributeError: 'NoneType' object has no attribute 'name'
```

justvanrossum:
If I start typing a string, a closing quote is added after the cursor. When I then manually type a closing quote, another set of quotes is added.

Still not convinced of this PR.

justvanrossum:
@andyclymer 

I think either the current quote behavior should be thoroughly fixed, or just dropped. Maybe it's good enough to just have automatic bracket/paren/brace closing. As long as:

- If one types an arbitrary sequence of characters blindly, one should always get exactly that sequence of characters. Any deviation of that is just going to be annoying and counter productive.

typemytype:
please re-rererereveiw :)

justvanrossum:
So far no improvement:

```
2018-03-30 11:24:23.800 DrawBotPy3[57314:5135078] PyObjC: Converting exception to Objective-C:
2018-03-30 11:24:23.800 DrawBotPy3[57314:5135078] Traceback (most recent call last):
2018-03-30 11:24:23.801 DrawBotPy3[57314:5135078]   File "/Users/just/code/git/drawbot/drawBot/ui/codeEditor.py", line 593, in keyDown_
2018-03-30 11:24:23.803 DrawBotPy3[57314:5135078]     
2018-03-30 11:24:23.803 DrawBotPy3[57314:5135078] if languageData:
2018-03-30 11:24:23.804 DrawBotPy3[57314:5135078] 
2018-03-30 11:24:23.805 DrawBotPy3[57314:5135078] NameError
2018-03-30 11:24:23.806 DrawBotPy3[57314:5135078] : 
2018-03-30 11:24:23.806 DrawBotPy3[57314:5135078] name 'languageData' is not defined
2018-03-30 11:24:23.807 DrawBotPy3[57314:5135078] 
2018-03-30 11:24:23.808 DrawBotPy3[57314:5135078] <class 'NameError'>: name 'languageData' is not defined
```

I can't type anything at all in the current state...

typemytype:
fixed

drawBot/ui/codeEditor.py
```diff
@@ -630,7 +630,7 @@ def keyDown_(self, event):
                         # jump the cursor after a quote
                         self.setSelectedRange_((selectedRange.location + 1, 0))
                         return
-                    if prevChar and selectedRange.length == 0:
+                    if prevChar and (prevChar not in autoCloseMap.keys() or prevChar in "\"'") and selectedRange.length == 0:
```
justvanrossum: `.keys()` is redundant.

typemytype:
I guess this is big enough for a fresh build

justvanrossum:
Perhaps we want to integrate #188 as well, but that's a tricky one.

PR #187, opened 2018-02-16
Memoize

typemytype:
based on Andy's PR see https://github.com/typemytype/drawbot/pull/185

but where the cache can be cleared for each context

### DIFF
--- a/drawBot/context/tools/openType.py
+++ b/drawBot/context/tools/openType.py
@@ -1,5 +1,7 @@
 import CoreText
 
+from drawBot.misc import memoize
+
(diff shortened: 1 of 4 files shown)


anthrotype:
i guess you know there's also https://docs.python.org/3/library/functools.html#functools.lru_cache

(oh.. too bad it's python 3 only)

anthrotype:
ah, but there's a backport!
https://github.com/jaraco/backports.functools_lru_cache
https://pypi.python.org/pypi/backports.functools_lru_cache

typemytype:
oh did not know...

there is one thing super specific for drawBot is that the cached functions must be cleared for each context... 

fe fonts are installed on the fly often...

thanks

PR #188, opened 2018-02-16
make the PY3 version the default one!

typemytype:
when this PR got approved I need to build an app first otherwise the links in the docs will not work correctly


justvanrossum:
I wonder how the update check/download feature will behave after this change?

typemytype:
good catch I have to switch these too https://github.com/typemytype/drawbot/blob/master/drawBot/updater.py#L56-L59

PR #211, opened 2018-06-07
FormattedString clear method onlyText argument

typoman:
It would be convenient if it was possible to only clear the text inside the attrubtedString and keep all the format styles.

### DIFF
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -902,8 +902,11 @@ class FormattedString(object):
         result.update(attributes)
         return result
 
-    def clear(self):
-        self._attributedString = AppKit.NSMutableAttributedString.alloc().init()
+    def clear(self, onlyText=False):
+        if onlyText:
+            self._attributedString.replaceCharactersInRange_withString_((0, self._attributedString.length() - 1), "")
+        else:
+            self._attributedString = AppKit.NSMutableAttributedString.alloc().init()
 
     def append(self, txt, **kwargs):
         """
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -904,7 +904,7 @@ class FormattedString(object):
 
     def clear(self, onlyText=False):
         if onlyText:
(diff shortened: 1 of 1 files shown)


drawBot/context/baseContext.py
```diff
@@ -902,8 +902,11 @@ def _validateAttributes(self, attributes, addDefaults=True):
         result.update(attributes)
         return result
 
-    def clear(self):
-        self._attributedString = AppKit.NSMutableAttributedString.alloc().init()
+    def clear(self, onlyText=False):
+        if onlyText:
+            self._attributedString.replaceCharactersInRange_withString_((0, self._attributedString.length()), "")
```
typemytype: I dont understand why there is a difference with `onlyText` keyword. Could you elaborate a use case for both?

and what would be the difference with creating a fresh FormatteString object? 
keeping the formatting is the only benefit, but then the internal attributedString just has to reset

typoman: The only advantage is I can copy a formattedString instance then clear the text but keep the style. It happens quite often that I want to start a new formatted string and I want to keep the style from previous one and this helps.

1. textOnly=True:
If I add a text to the object it keeps the style from what I set before so I don't have to repeat adding the style.

2. textOnly=False:
Everything gets reset so it's like having a new fresh formatted string.

typemytype: the current implementation of a `formattedString.clear()` removes already all the text from the string an keeps the formatting:

```python
t = FormattedString()

t.font("Times-Italic")

t += "hello world"
print(len(t))
print(t._font)
t.clear()
print(len(t))
print(t._font)
```

the `onlyText` keyword is removing all the text from the internal attributedString object, whenever a fromattedString object adds text it will create attributes for the added text.

typoman: When I test it now they style is kept! I can't reproduce the problem I had before so I think you can ignore my pull request.

typemytype: thanks for checking!!

Issue #215, opened 2018-07-06
[proposal] add imageResolution(path) 


typemytype:
return the dpi of the image, similar as what can be used in  `saveImage`

see http://forum.drawbot.com/topic/66/how-to-get-image-dpi

typemytype:
can an image have different resolution for width and height? @justvanrossum 

```python
path = 'pathToImage.png'

import AppKit

url = AppKit.NSURL.fileURLWithPath_(path)

# im = AppKit.NSImage.alloc().initByReferencingURL_(url)
# rep = im.representations()[0]

rep = AppKit.NSImageRep.imageRepWithContentsOfURL_(url)

print(rep.pixelsWide() / rep.size().width * 72)
print(rep.pixelsHigh() / rep.size().height * 72)
```

PR #224, opened 2018-09-25
Image resolution

typemytype:
see #215 

todo:

* write docs
* add tests


drawBot/drawBotDrawingTools.py
```diff
@@ -1978,6 +1978,33 @@ def imagePixelColor(self, path, xy):
...
+    def imageResolution(self, path):
+        """
+        Return the image resolution for a given image.
+        """
+        if isinstance(path, AppKit.NSImage):
+            # its an NSImage
+            # get the bitmap representation
+            try:
+                rep = path.representations()[0]
+            except Exception:
```
justvanrossum: Isn't it a simple IndexError you want to catch here?

drawBot/drawBotDrawingTools.py
```diff
@@ -1978,6 +1978,33 @@ def imagePixelColor(self, path, xy):
...
+                path = optimizePath(path)
+            if path.startswith("http"):
+                url = AppKit.NSURL.URLWithString_(path)
+            else:
+                if not os.path.exists(path):
+                    raise DrawBotError("Image does not exist")
+                url = AppKit.NSURL.fileURLWithPath_(path)
+                try:
+                    rep = AppKit.NSImageRep.imageRepWithContentsOfURL_(url)
+                except Exception:
```
justvanrossum: Can this exception be more specific, too?

typemytype: this catches a range of errors:  file not found, file not readable, file not valid, ...

drawBot/drawBotDrawingTools.py
typemytype: fixed

PR #235, opened 2018-10-28
Attempts to improve point pen doco

justvanrossum:
Maybe leave this open a little while, in case more can be improved.


justvanrossum:
Or just merge if you agree with this wording :)

PR #243, opened 2018-11-28
support for .icns files

typemytype:


typemytype:
still todo: write test

drawBot/context/icnsContext.py
```diff
@@ -0,0 +1,60 @@
...
+
+class ICNSContext(ImageContext):
+
+    fileExtensions = ["icns"]
+
+    allowedPageSizes = [16, 32, 128, 256, 512, 1024]
+
+    def _writeDataToFile(self, data, path, options):
+        # create a iconset folder
+        iconsetPath = tempfile.mkdtemp(suffix=".iconset")
```
justvanrossum: The rest of the code should be in a try/finally, nw there's a chance this temp folder won't get deleted. Better yet, use `tempfile.TemporaryDirectory()` as a context manager, if Py3-only is ok.

drawBot/context/icnsContext.py
```diff
@@ -0,0 +1,60 @@
...
+        # get the complete pdf
+        pdfDocument = Quartz.PDFDocument.alloc().initWithData_(data)
+        pageCount = pdfDocument.pageCount()
+        # set the image resolution
+        options["imageResolution"] = 72
+        # make a copy and alter the resolution
+        options_2x = dict(options)
+        options_2x["imageResolution"] = 144
+        # start loop over all pages
+        for index in range(0, pageCount):
```
justvanrossum: `range(pageCount)` will do the same.

drawBot/context/icnsContext.py
```diff
@@ -0,0 +1,60 @@
...
+            # get the pdf page
+            page = pdfDocument.pageAtIndex_(index)
+            # get the pdf page, this acts also as pdf document...
+            pageData = page.dataRepresentation()
+            # extract the size of the page
+            _, (w, h) = page.boundsForBox_(Quartz.kPDFDisplayBoxArtBox)
+            w = int(round(w))
+            h = int(round(h))
+            # dont allow any other size, the command iconutil will not work otherwise
+            if w not in self.allowedPageSizes or h not in self.allowedPageSizes:
```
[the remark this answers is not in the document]
justvanrossum: Or: `if w not in self.allowedPageSizes or w != h:`

drawBot/context/icnsContext.py
typemytype: we have to setup a plan how long py2 is been supported, it is for now...

PR #255, opened 2019-01-15
trying to improve text width

### LINKED ISSUE #254: Line wrap occurring for non-latin scripts
Hello,

I would like to report a bug which causes line wraps when there are no width constraints. This seems to be a script related issues because I only encounter it with non-latin scripts (i.e. Armenian, Cyrillic). 

I've attached a drawbot script and font file which illustrates this issue. 

Note: The font file is simplified variable font with three scripts ( Latin, Armenian, and Cyrillic). For simplicity, there is only one letter per script ('N', 'Ո', 'П' and each letter uses the same rectangle with varying widths across 3 masters. 

screenshot 2019-01-11 at 14 23 10

[drawbot_script_test.zip](https://github.com/typemytype/drawbot/files/2751173/drawbot_script_test.zip)

typemytype:
an nsattributestring does not provide a correct w, h for a CTFrameSetter

fixes #254

### DIFF
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -272,6 +272,7 @@ class BezierPath(BasePen):
 
         attributedString = context.attributedString(txt, align)
         w, h = attributedString.size()
+        w *= 2
         if offset:
             x, y = offset
         else:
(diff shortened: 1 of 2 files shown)


justvanrossum:
How is `textSize()` affected by this issue?

justvanrossum:
This PR should be reverted, as it caused #265.

Issue #256, opened 2019-01-21
missing varfont api: listInstances


typemytype:
requested here https://forum.drawbot.com/topic/121/listinstances-font

typemytype:
this should do it

```python
import CoreText
import AppKit

from fontTools.ttLib import TTFont

from drawBot.context.tools import variation


def getIntancesForFontName(fontName):
    instances = {}
    font = CoreText.CTFontCreateWithName(fontName, 12, None)
    if font is None:
        return instances
    cgFont = CoreText.CGFontCreateWithFontName(fontName)
    fontDescriptor = font.fontDescriptor()
    url = CoreText.CTFontDescriptorCopyAttribute(fontDescriptor, CoreText.kCTFontURLAttribute)
    if url is None:
        return instances

    variationAxesDescriptions = CoreText.CTFontCopyVariationAxes(font)
    tagNameMap = {}
    for variationAxesDescription in variationAxesDescriptions:
        tag = variation.convertIntToVariationTag(variationAxesDescription[CoreText.kCTFontVariationAxisIdentifierKey])
        name = variationAxesDescription[CoreText.kCTFontVariationAxisNameKey]
        tagNameMap[tag] = name

    ft = TTFont(url.path())
    if "fvar" in ft:
        fvar = ft["fvar"]

        for instance in fvar.instances:
            fontVariations = dict()
            for axis, value in instance.coordinates.items():
                fontVariations[tagNameMap[axis]] = value

            varFont = CoreText.CGFontCreateCopyWithVariations(cgFont, fontVariations)   
            postScriptName = CoreText.CGFontCopyPostScriptName(varFont)
            instances[postScriptName] = instance.coordinates
            
    ft.close()
    return instances

r = getIntancesForFontName("Skia")
print(r)
```

justvanrossum:
This has been implemented in #259.

PR #259, opened 2019-01-24
Named instances

typemytype:
fixing #256

### DIFF
--- a/docs/content/text/textProperties.rst
+++ b/docs/content/text/textProperties.rst
@@ -12,6 +12,7 @@ Text Properties
 .. autofunction:: drawBot.listOpenTypeFeatures
 .. autofunction:: drawBot.fontVariations(wdth=0.6, wght=0.1, ...)
 .. autofunction:: drawBot.listFontVariations
+.. autofunction:: drawBot.listFontVariationNamedInstances
 .. autofunction:: drawBot.tabs
 .. autofunction:: drawBot.language
 
--- a/docs/content/text/textProperties.rst
+++ b/docs/content/text/textProperties.rst
@@ -12,7 +12,7 @@ Text Properties
 .. autofunction:: drawBot.listOpenTypeFeatures
 .. autofunction:: drawBot.fontVariations(wdth=0.6, wght=0.1, ...)
 .. autofunction:: drawBot.listFontVariations
-.. autofunction:: drawBot.listFontVariationNamedInstances
+.. autofunction:: drawBot.listNamedInstances
 .. autofunction:: drawBot.tabs
 .. autofunction:: drawBot.language
 
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -1396,6 +1396,9 @@ class FormattedString(object):
         """
         List all variation axes for the current font.
 
+        Returns a dictionary with all axis tags instance with an info dictionary with the following keys: `name`, `minValue` and `maxValue`.
+        For non variable fonts an empty dictionary is returned.
+
         Optionally a `fontName` can be given. If a font path is given the font will be installed and used directly.
         """
         if fontName:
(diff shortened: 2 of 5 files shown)


docs/content/text/textProperties.rst
```diff
@@ -12,6 +12,7 @@ Text Properties
 .. autofunction:: drawBot.listOpenTypeFeatures
 .. autofunction:: drawBot.fontVariations(wdth=0.6, wght=0.1, ...)
 .. autofunction:: drawBot.listFontVariations
+.. autofunction:: drawBot.listFontVariationNamedInstances
```
justvanrossum: I wonder if `listNamedInstances()` would be good enough?

drawBot/context/baseContext.py
```diff
@@ -1404,6 +1404,18 @@ def listFontVariations(self, fontName=None):
             fontName = self._font
         return variation.getVariationAxesForFontName(fontName)
 
+    def listFontVariationNamedInstances(self, fontName=None):
+        """
+        List all named instances for the current font.align
```
justvanrossum: Align? Also, mention variable fonts here, and what it will return if the font is no variable font or has no named instance.

drawBot/context/tools/variation.py
```diff
@@ -42,3 +44,45 @@ def getVariationAxesForFontName(fontName):
...
+@memoize
+def getVariationNamedInstancesForFontName(fontName):
+    """
+    Return a dict { postscriptName: location } of all named instances in a given font.
+    """
+    instances = {}
+    font = CoreText.CTFontCreateWithName(fontName, 12, None)
+    if font is None:
+        return instances
+    cgFont = CoreText.CGFontCreateWithFontName(fontName)
```
justvanrossum: `cgFont` is only used in the `if "fvar" in ft` code path, better move it there

drawBot/context/tools/variation.py
```diff
@@ -42,3 +44,45 @@ def getVariationAxesForFontName(fontName):
...
+    fontDescriptor = font.fontDescriptor()
+    url = CoreText.CTFontDescriptorCopyAttribute(fontDescriptor, CoreText.kCTFontURLAttribute)
+    if url is None:
+        return instances
+
+    variationAxesDescriptions = CoreText.CTFontCopyVariationAxes(font)
+    if variationAxesDescriptions is None:
+        # 'normal' fonts have no named instances
```
justvanrossum: "non-variable fonts"

drawBot/drawBotDrawingTools.py
```diff
@@ -1560,6 +1560,14 @@ def listFontVariations(self, fontName=None):
         """
         return self._dummyContext._state.text.listFontVariations(fontName)
 
+    def listFontVariationNamedInstances(self, fontName=None):
+        """
+        List all named instances for the current font.align
```
justvanrossum: Can the doc string somehow be shared with baseContext.py?

docs/content/text/textProperties.rst
typemytype: also fine for me: named instances only appear within var fonts (that the only reason why I added the long method name

drawBot/drawBotDrawingTools.py
typemytype: I added now a couple of them

this needs to happen for other methods docs strings.

drawBot/context/baseContext.py
typemytype: typo...

docs/content/text/textProperties.rst
justvanrossum: This is still the old name.

drawBot/context/tools/variation.py
typemytype: this is solved

typemytype:
this is a nice new feature for a release

PR #278, opened 2019-06-18
raise when text is no string or formattedstring

### LINKED ISSUE #276: Confusing traceback when wrong data type is used for text
(Found out in the context of a workshop)

When using an integer in a text field, the traceback is too complicated:
```
for n in range(10):
    page_number = n + 1
    newPage(100, 100)
    text(page_number, (10, 10))
```

```
/Applications/DrawBot.app/Contents/Resources/lib/python36.zip/drawBot/context/baseContext.py:1134: UninitializedDeallocWarning: leaking an uninitialized object of type NSConcreteAttributedString
  File "<untitled>", line 4, in <module>
  File "drawBot/drawBotDrawingTools.pyc", line 1597, in text
  File "drawBot/context/baseContext.pyc", line 2209, in attributedString
  File "drawBot/context/baseContext.pyc", line 1134, in append
ValueError: NSInvalidArgumentException - -[OC_BuiltinPythonNumber length]: unrecognized selector sent to instance 0x6000000b6c00
```

It would be cool (and less scary for students) to say something like “Text needs to be a string”.
Thanks!

typemytype:
fixes #276

### DIFF
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -265,6 +265,8 @@ class BezierPath(BasePen):
 
         Optionally `txt` can be a `FormattedString`.
         """
+        if not isinstance(txt, (basestring, FormattedString)):
+            raise ValueError("depythonifying 'str' or 'FormattedString', got '%s'" % type(txt).__name__)
         if align and align not in BaseContext._textAlignMap.keys():
             raise DrawBotError("align must be %s" % (", ".join(BaseContext._textAlignMap.keys())))
         context = BaseContext()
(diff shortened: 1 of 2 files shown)


drawBot/context/baseContext.py
```diff
@@ -265,6 +265,8 @@ def text(self, txt, offset=None, font=_FALLBACKFONT, fontSize=10, align=None):
 
         Optionally `txt` can be a `FormattedString`.
         """
+        if not isinstance(txt, (basestring, FormattedString)):
+            raise ValueError("depythonifying 'str' or 'FormattedString', got '%s'" % type(txt).__name__)
```
justvanrossum: Should be `TypeError`, and the message should say "expected 'str' or 'FormattedString', got '%s' ..."

typemytype: Made ValueError as what this is what pyobjc is raising... TypeError is idd better

justvanrossum:
Looks good to me.

PR #279, opened 2019-06-28
add fallbackfont into svg

### LINKED ISSUE #274: svg fallback fonts
use the fallback font, if set as fallback in svg

requested here: https://forum.drawbot.com/topic/145/setting-fallback-font-for-text-in-svg

typemytype:
Fixes #274


typemytype:
still todo: add test

typemytype:
tests have to be review in general... as they fail massively based on the OS version

justvanrossum:
Still, a test for this would be good. And I doubt this particular test will fall into the category that depends on the OS version.

Other than that: looking good.

typemytype:
test added!

PR #282, opened 2019-07-01
adding arcTo example script

gferreira:
example taken from [arcTo documentation? (DB Forum)](https://forum.drawbot.com/topic/100/arcto-documentation/6)


typemytype:
thanks! nice example

PR #289, opened 2019-08-15
Remove support for Python 2

justvanrossum:
Mostly to get rid of usages of `fontTools.misc.py23.PY2` and `fontTools.misc.py23.PY3`, but went ahead and removed most usage of py23 altogether.

Todo:
- [x] `__future__` statements
- [x] `py23` usage in tests
- [x] `__future__` flag usage in code compilation
- [x] get rid of `__future__` division preference

### DIFF
--- a/DrawBot.py
+++ b/DrawBot.py
@@ -5,8 +5,6 @@ import sys
 import os
 import random
 
-from fontTools.misc.py23 import PY3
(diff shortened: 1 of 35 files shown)


typoman:
There are still traces of that in the setup.

PR #307, opened 2019-10-09
CTFeatureOpenTypeTag

typemytype:


typemytype:
to problem still to solve:

1) what is not possible anymore is to get the default settings for opentype feature tags

     but I guess CoreText is using the recommendations listed **somewhere** 
there are some here https://docs.microsoft.com/en-us/typography/opentype/spec/features_ae in the `UI suggestion` part...

2) for core macOS fonts like (Helvetica.ttc, Hoefler Text.ttc) we need to parse the `feat` table instead according this: https://developer.apple.com/fonts/TrueType-Reference-Manual/RM09/AppendixF.html

@jjgod @folengo

typemytype:
this works now on 10.15

but no opentype features for fonts with a `feat` table

a direct download can be dowloaded here: http://static.typemytype.com/drawBot/DrawBot10.15.dmg

typemytype:
fonts with `feat` should be supported now

justvanrossum:
The output of `listOpenTypeFeatures()` is not backwards compatible.

drawBot/context/baseContext.py
```diff
@@ -1777,7 +1780,7 @@ def appendGlyph(self, *glyphNames):
         # disable calt features, as this seems to be on by default
         # for both the font stored in the nsGlyphInfo as in the replacement character
         fontAttributes = {}
-        fontAttributes[CoreText.NSFontFeatureSettingsAttribute] = [openType.featureMap["calt_off"]]
+        fontAttributes[CoreText.kCTFontFeatureSettingsAttribute] = [dict(CTFeatureOpenTypeTag="calt", CTFeatureOpenTypeValue=False)]
```
typemytype: I guess if this needs to work on 10.10 the olde style font attributes must be added too.

justvanrossum: Hmm, this oddly enough seems to work as expected on 10.10.

typemytype:
cool! building a fresh build later on today

PR #308, opened 2019-10-13
add methods to fit and center the bezierpath to a given bounds

typoman:
I've used these many times and I thought it could be useful to be added to the BezierPath object. I hope it's not too arbitrary!


typoman:
Also if you think this can be added, I think alignTop, alignBottom, etc to a given bound is also useful.

typoman:
test case:
```
from drawBot import *
from datetime import datetime


size(800, 800)
b = BezierPath()
b.oval(234, 324, 345, 342)
fill(None)
stroke(0)

b2 = BezierPath()
b2.rect(200, 200, 400, 400)
b.fit(b2.bounds())
drawPath(b2)
drawPath(b)


import os, subprocess
fn = os.path.split(os.path.realpath(__file__))[1] + datetime.now().strftime("%Y|%m|%d-%H%M%S")
saveImage(f"{fn}.pdf")
subprocess.call(['open', "%s.pdf" %fn])
```

typoman:
After a bit of thinking, maybe align to a path is better than bounds. Then there can be a method as alignToPath(other_path, "center") which does the job?

justvanrossum:
> After a bit of thinking, maybe align to a path is better than bounds.

Hm, I don't think so. I want to think a bot more about this PR. The base idea is good, and I've used similar things (but not exactly the same :) ) a lot before, so I agree something like this is useful.

drawBot/context/baseContext.py
```diff
@@ -537,7 +537,34 @@ def transform(self, transformMatrix, center=(0, 0)):
         aT.setTransformStruct_(transformMatrix[:])
         self._path.transformUsingAffineTransform_(aT)
 
-    # boolean operations
```
justvanrossum: This comment should stay, but below your new methods.

drawBot/context/baseContext.py
```diff
@@ -537,7 +537,34 @@ def transform(self, transformMatrix, center=(0, 0)):
...
+        b_w = b_x2 - b_x1
+        b_h = b_y2 - b_y1
+        if w / h >= b_w / b_h:
+            scale = b_w / w
+        else:
+            scale = b_h / h
+        self.scale(scale, scale)
+        self.center(bounds)
+
+    def center(self, bounds):
```
justvanrossum: I propose the following instead of center:
```python
    def align(self, point, horizontalAlignment=None, verticalAlignment=None):
        ...horizontalAlignment can be None, "center", "left" or "right"
        ...verticalAlignment can be None, "center", "top" or "bottom"
        ...`point` is an (x, y) tuple, to be used as the alignment target.
        ...for example bez.align((100, 200), "center", "bottom") will align the 
           bounding box of the shape to be centered horizontally at 100,
           and the bottom with 200. An alignment value of None will make
           the path not move in that direction.
```

drawBot/context/baseContext.py
```diff
@@ -537,7 +537,34 @@ def transform(self, transformMatrix, center=(0, 0)):
         aT.setTransformStruct_(transformMatrix[:])
         self._path.transformUsingAffineTransform_(aT)
 
-    # boolean operations
+    def fit(self, bounds):
```
justvanrossum: I propose to extend this:
```python
    def fit(self, bounds, constrain=True):
        ...
```
Where `constrain=True` behaves like what you have implemented, and `False` would mean: fill up the whole bounds.

Maybe `bounds` should be called `box`, and have the same meaning as in `textBox()`: `(x, y, w, h)`, as this is more common in DrawBot overall.

justvanrossum: Or perhaps `constrainProportions`.

drawBot/context/baseContext.py
typoman: Oops, I will fix that.

drawBot/context/baseContext.py
typoman: I agree with box instead of bounds. I used bounds because it can be used to align one shape to another, but it could also be confusing.

Could you please elaborate on fill up bounds, I didn't get the idea.

drawBot/context/baseContext.py
typoman: Nice, that's more versatile! Will fix.

drawBot/context/baseContext.py
justvanrossum: You now scale by a factor that will make the shape fit in the box, while maintaining widht/height proportions (constrain). You could also scale x and y independently, so the bounding box of the shape will fill the entire target box. That's what I mean with `constrainProportions=False`. Hope that clears it up.

typoman: Ow I see! Like keeping the ratio.

justvanrossum:
Are you interested in continuing/finishing this PR? If not, we should probably close.

PR #309, opened 2019-10-14
[WIP] travis support second attempt

typemytype:


justvanrossum:
Sorry for the weird history: I messed up rebasing from master... The result should be ok.

typemytype:
I would prefer to use a macOS system font instead of ACaslon

PR #310, opened 2019-10-20
expandStroke method for BezierPath object

typoman:
~~The method strokes the path itself, so it does not return anything~~. I changed the _getCGPath so it doesn't add the bounds when it's not needed and moved the _LINEJOINSTYLESMAP and _LINECAPSTYLESMAP outside the BaseContext. Not sure if it suits the clean look of the code.

### DIFF
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -393,7 +393,7 @@ class BezierPath(BasePen):
         """
         return self._path
 
-    def _getCGPath(self, withBounds=False):
+    def _getCGPath(self):
         path = Quartz.CGPathCreateMutable()
         count = self._path.elementCount()
         for i in range(count):
@@ -411,14 +411,6 @@ class BezierPath(BasePen):
                 )
             elif instruction == AppKit.NSClosePathBezierPathElement:
                 Quartz.CGPathCloseSubpath(path)
-        # hacking to get a proper close path at the end of the path
-        if withBounds:
-            x, y, _, _ = self.bounds()
-            Quartz.CGPathMoveToPoint(path, None, x, y)
-            Quartz.CGPathAddLineToPoint(path, None, x, y)
-            Quartz.CGPathAddLineToPoint(path, None, x, y)
-            Quartz.CGPathAddLineToPoint(path, None, x, y)
-            Quartz.CGPathCloseSubpath(path)
         return path
 
     def _setCGPath(self, cgpath):
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -14,7 +14,16 @@ from .tools import variation
 
 
 _FALLBACKFONT = "LucidaGrande"
-
+_LINEJOINSTYLESMAP = dict(
+    miter=Quartz.kCGLineJoinMiter,
+    round=Quartz.kCGLineJoinRound,
+    bevel=Quartz.kCGLineJoinBevel
+)
+_LINECAPSTYLESMAP = dict(
+    butt=Quartz.kCGLineCapButt,
+    square=Quartz.kCGLineCapSquare,
+    round=Quartz.kCGLineCapRound,
+)
 
 def _tryInstallFontFromFontName(fontName):
     from drawBot.drawBotDrawingTools import _drawBotDrawingTool
@@ -384,7 +393,7 @@ class BezierPath(BasePen):
         """
         return self._path
 
-    def _getCGPath(self):
+    def _getCGPath(self, withBounds=False):
         path = Quartz.CGPathCreateMutable()
         count = self._path.elementCount()
         for i in range(count):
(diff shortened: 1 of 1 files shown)


drawBot/context/baseContext.py
```diff
@@ -2187,16 +2204,16 @@ def miterLimit(self, value):
...
+        if join not in _LINEJOINSTYLESMAP:
             raise DrawBotError("lineJoin() argument must be 'bevel', 'miter' or 'round'")
-        self._state.lineJoin = self._lineJoinStylesMap[join]
+        self._state.lineJoin = _LINEJOINSTYLESMAP[join]
 
     def lineCap(self, cap):
         if cap is None:
             self._state.lineCap = None
-        if cap not in self._lineCapStylesMap:
+        if cap not in self._LINECAPSTYLESMAP:
```
justvanrossum: That `self.` should go.

drawBot/context/baseContext.py
```diff
@@ -403,14 +412,30 @@ def _getCGPath(self):
             elif instruction == AppKit.NSClosePathBezierPathElement:
                 Quartz.CGPathCloseSubpath(path)
         # hacking to get a proper close path at the end of the path
-        x, y, _, _ = self.bounds()
-        Quartz.CGPathMoveToPoint(path, None, x, y)
-        Quartz.CGPathAddLineToPoint(path, None, x, y)
-        Quartz.CGPathAddLineToPoint(path, None, x, y)
-        Quartz.CGPathAddLineToPoint(path, None, x, y)
-        Quartz.CGPathCloseSubpath(path)
+        if withBounds:
```
justvanrossum: It's not drawing bounds, really, just adding the same point several times over. I don't understand the purpose, and I hope it can be deleted. @typemytype?

drawBot/context/baseContext.py
```diff
@@ -620,6 +645,10 @@ def intersectionPoints(self, other=None):
             contours += other._contoursForBooleanOperations()
         return booleanOperations.getIntersections(contours)
 
+    def strokePath(self, width, lineCap="round", lineJoin="round", miterLimit=10):
```
justvanrossum: Needs a good doc string.

drawBot/context/baseContext.py
typemytype: Ive I remember it correctly a text setter needs a proper closed path... will test again...

typemytype: Just could you run the test without adding that single point?

it was added here https://github.com/typemytype/drawbot/commit/9f48f99bfc0dffa70ab759a42a2ea5fa70f985ab
while working on the hyphenation...

drawBot/context/baseContext.py
```diff
@@ -403,14 +412,30 @@ def _getCGPath(self):
...
+        if withBounds:
+            x, y, _, _ = self.bounds()
+            Quartz.CGPathMoveToPoint(path, None, x, y)
+            Quartz.CGPathAddLineToPoint(path, None, x, y)
+            Quartz.CGPathAddLineToPoint(path, None, x, y)
+            Quartz.CGPathAddLineToPoint(path, None, x, y)
+            Quartz.CGPathCloseSubpath(path)
         return path
 
+    def _setCGpath(self, cgpath):
```
typemytype: the camelcase method name should be `_setCGPath`

drawBot/context/baseContext.py
```diff
@@ -403,14 +412,30 @@ def _getCGPath(self):
...
+        self._path = AppKit.NSBezierPath.alloc().init()
+
+        def _addPoints(arg, element):
+            instruction, points = element.type, element.points
+            if instruction == Quartz.kCGPathElementMoveToPoint:
+                self._path.moveToPoint_((points[0].x, points[0].y))
```
typemytype: the points are already valid nsBezierPath points
you can provide the point directly: `moveToPoint_(points[0])` same for `lineToPoint_` and `curveToPoint_controlPoint1_controlPoint2_`

drawBot/context/baseContext.py
typoman: Could you please check if now it's good?

drawBot/context/baseContext.py
typoman: I leave it up to you ( ͡~ ͜ʖ ͡°)

drawBot/context/baseContext.py
typemytype: maybe a different method name?

* `strokedPath`
* `expandPath`

strokePath is to ambiguous, it does not draw a stroked path...

drawBot/context/baseContext.py
typoman: just removed it, thanks!

drawBot/context/baseContext.py
```diff
@@ -620,6 +645,10 @@ def intersectionPoints(self, other=None):
             contours += other._contoursForBooleanOperations()
         return booleanOperations.getIntersections(contours)
 
+    def strokePath(self, width, lineCap="round", lineJoin="round", miterLimit=10):
+        strokedCGPath = Quartz.CGPathCreateCopyByStrokingPath(self._getCGPath(), None, width, _LINECAPSTYLESMAP[lineCap], _LINEJOINSTYLESMAP[lineJoin], miterLimit)
+        self._setCGpath(strokedCGPath)
```
typemytype: I think it should return fresh bezierPath object with a stroked path and not add it to itself. 
This is a similar action like the boolean operation actions

typoman: Yeah sounds better!

drawBot/context/baseContext.py
typoman: Thanks!

drawBot/context/baseContext.py
typoman: fixed

drawBot/context/baseContext.py
typoman: `expandPath` sounds better maybe, what do you think?

typoman: for the docstring how about this:
```
        Returns an expanded stroke around the path with the given width.

        Optionally these options can be set:

        * `lineCap`: Possible values are `"butt"`, `"square"` or `"round"`
        * `lineJoin`: Possible values are `"bevel"`, `"miter"` or `"round"`
        * `mitterLimit`: The miter limit to use for `"miter"` lineJoin option
```

typemytype: something like this: `Returns a new bezier path with an expanded stroke around the bezier path with a given width.`

typoman: sounds good. How about `expandStoke` for the method name?

typemytype: +1 for `expandStroke`

drawBot/context/baseContext.py
justvanrossum: The tests run fine without that section.

justvanrossum: So the whole `withBounds` parameter thing can go.

typoman: Just pushed!

drawBot/context/baseContext.py
```diff
@@ -620,6 +637,27 @@ def intersectionPoints(self, other=None):
...
+    def expandStroke(self, width, lineCap="round", lineJoin="round", miterLimit=10):
+        """
+        Returns a new bezier path with an expanded stroke around the bezier
+        path with a given width.
+
+        Optionally these options can be set:
+
+        * `lineCap`: Possible values are `"butt"`, `"square"` or `"round"`
+        * `lineJoin`: Possible values are `"bevel"`, `"miter"` or `"round"`
+        * `mitterLimit`: The miter limit to use for `"miter"` lineJoin option
```
justvanrossum: It says `mitterLimit` instead of `miterLimit` here.

justvanrossum: Also, needs some rewording:

```python
"""
Returns a new bezier path with an expanded stroke around the original path,
with a given `width`. Note: the new path will not contain the original path.

The following optional arguments are available with respect to line caps and joints:
* `lineCap`: Possible values are `"butt"`, `"square"` or `"round"`
* `lineJoin`: Possible values are `"bevel"`, `"miter"` or `"round"`
* `miterLimit`: The miter limit to use for `"miter"` lineJoin option
"""
```

justvanrossum: Sorry, "joins", not "joints"

typoman: Thank you Just and sorry for my bad grammer! ( ͡~ ͜ʖ ͡°)

justvanrossum:
Can you update the title of this PR te reflect the new method name?

typoman:
Of course!

typemytype:
Thanks!!

typoman:
Anytime!

PR #311, opened 2019-10-22
Changes text() argument name to "position" instead of "x"

andyclymer:
Improves the error message when an argument is missing in the `text()` function. This came up during a class I was teaching, without any arguments this is the error message —

`TypeError: text() missing 2 required positional arguments: 'txt' and 'x'`

...which isn't clear, since `x` should technically be a tuple of `(x, y)`. I understand that it's named this way for legacy reasons (which is why there's an optional argument for `y`) but it would be more clear to rename the argument as `position` —

`TypeError: text() missing 2 required positional arguments: 'txt' and 'position'`

This is the same argument name that's used in `image()` when it also needs a tuple for position.

### DIFF
--- a/drawBot/drawBotDrawingTools.py
+++ b/drawBot/drawBotDrawingTools.py
@@ -1577,7 +1577,7 @@ class DrawBotDrawingTool(object):
 
     # drawing text
 
-    def text(self, txt, x, y=None, align=None):
+    def text(self, txt, position, y=None, align=None):
         """
         Draw a text at a provided position.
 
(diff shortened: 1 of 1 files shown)


typemytype:
+1 to remove these super old argument usage

It happens in lots of places

justvanrossum:
Fixed in https://github.com/typemytype/drawbot/commit/d96d7bf136606a49a68b4b752163219a86d627ba (I botched the PR number in the commit). Thanks Andy!

Issue #314, opened 2019-10-28
Some test suite items need fixing/updating


justvanrossum:
- [ ] Some examples use ACaslon as a demo font. We should switch to a system font, eg. HoeflerText. See the examples named `formattedString`, `openTypeFeatures` and `openTypeFeaturesFormattedString`
- [ ] The `lineHeight` example mysteriously fails badly on Travis
- [ ] testExport.py/ExportTest.test_imageFallbackBackgroundColor() uses excessive rounding in the last test, because somehow `NSColor.redColor()` does not translate exactly to a `(1, 0, 0)` pixel color
- [ ] One of the VF demos using Skia fail in subtle ways: `fontVariations2.py`
- [ ] `http` images are not supported on Travis. `https` _should_ work, but our `http` image URL doesn't work as `https`. (We mock `image()` in the example tests with a local image, but I think our `imageHTTP.py` test should be real.)

PR #316, opened 2019-10-28
replace ACaslon to Baskerville

typemytype:
Baskerville is a default macOS font, so must work in different setups

see #314


drawBot/context/baseContext.py
```diff
@@ -1391,7 +1391,7 @@ def openTypeFeatures(self, *args, **features):
             # create an empty formatted string object
             t = FormattedString()
             # set a font
-            t.font("ACaslonPro-Regular")
+            t.font("Didot")
```
justvanrossum: But this font does not have 'lnum'. The example should switch to 'onum' I think.

PR #317, opened 2019-10-28
Added codecov

yanone:
This should get you started. You may have to create an account at https://codecov.io using your Github account.
The "passing" badge is useless without measuring the code covered in the tests, because I can write a one-line test script and show a passing badge. But it's not for show-off. It shows you which parts of the code haven't been included in the testing yet.
With Type.World, I'm at 95% right now. Covering the few extra percent costs a lot of time because I basically have to construct artificial errors. But it's worth it. I find *so* many mistakes through CI & code coverage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
     osx_image: xcode9.4  # 10.13
     language: generic
     script:
-    - "./.travis/tests.sh"
+    - coverage run "./.travis/tests.sh"
 #   - os: osx
 #     osx_image: xcode9.4  # 10.13
 #     language: generic
@@ -69,3 +69,6 @@ matrix:
 # #   on:
 # #     repo: typemytype/drawbot
 # #     branch: trying_out_travis  # remove when PR is done
+
+after_success:
+  - codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
     osx_image: xcode9.4  # 10.13
     language: generic
     script:
-    - coverage run "./.travis/tests.sh"
+    - "./.travis/tests.sh"
 #   - os: osx
 #     osx_image: xcode9.4  # 10.13
 #     language: generic
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -16,4 +16,3 @@ fi
 command -v python3
 python3 --version
 python3 -c "import platform; print('macOS version:', platform.mac_ver()[0])"
-pip3 install codecov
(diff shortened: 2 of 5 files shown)


typemytype:
Ive adde drawBot to my codecov account 
do I have to add the token anywhere?

yanone:
I believe not. This should be handled automatically by Travis CI after the tests there have completed  successfully.

.travis.yml
```diff
@@ -17,7 +17,7 @@ matrix:
     osx_image: xcode9.4  # 10.13
     language: generic
     script:
-    - "./.travis/tests.sh"
+    - coverage run "./.travis/tests.sh"
```
justvanrossum: I think the `coverage run` bit should be inside `.travis/tests.sh`, probably replacing this line:

`python3 ./tests/runAllTests.py`

by

`coverage run ./tests/runAllTests.py`

typemytype:
so how does codecov knows about the coverage? so it can make that svg image

yanone:
I wish I could answer that. They somehow measure the code as it's being executed.

.travis/install.sh
```diff
@@ -16,3 +16,4 @@ fi
 command -v python3
 python3 --version
 python3 -c "import platform; print('macOS version:', platform.mac_ver()[0])"
+pip3 install codecov
```
justvanrossum: This should also move to the `tests.sh` script, where the other pip installs are being done.

(I'm not sure our current setup is good, but I don't think we can pip install codecov before our venv is activated.)

yanone: But you also can’t run the test script without codecov being installed: `coverage run "./.travis/tests.sh"`

yanone: Sorry, I didn't read your previous comment. Both ways should work.

justvanrossum: No, we shouldn't run coverage on pip install etc... The `pip install codecov` _and_ the  `coverage run ...` call should move to tests.sh.

(Also: `coverage` is a Python script runner. It doesn't work with shell scripts.)

yanone: Okay, I moved it like you said 🤞

justvanrossum:
Thanks Yanone! Looking good. I'll wait for the build to finish.

yanone:
It failed (internet issues). But I don't have permissions to restart it.

justvanrossum:
Hm, I can't do it either. We have to wait for @typemytype or maybe just push a harmless commit? Thanks.

justvanrossum:
It's running again!

Issue #315, opened 2019-10-28
Last character is sometimes dropped when using Variable Fonts


justvanrossum:
Depending on the axis settings, DrawBot sometimes drops the last letter from a string as set with `text()`.

@LettError 
Using `MutatorSans.ttf` font from https://github.com/LettError/mutatorSans

```python
txt = '→→→→'

values = [('ok', 500), ('dropping', 502)]
weightValue = 200

for tag, widthValue in values:    
    newPage(690, 240)
    
    font("MutatorMathTest")
    fontSize(200)
    fontVariations(wght=weightValue, wdth=widthValue)
    w, h = textSize(txt)
    fill(0,0.8,0.8,0.3)
    rect(20, 20, w, h)

    fill(0.8,0,0.8,0.3)
    text(txt, (20, 20), align="left")
    saveImage('drawbot_drops_letter_%s.jpg' % tag)
```

![drawbot_drops_letter_dropping](https://user-images.githubusercontent.com/4246121/67683122-0f132d80-f991-11e9-8a7c-06535e05f239.jpg)
![drawbot_drops_letter_ok](https://user-images.githubusercontent.com/4246121/67683134-15090e80-f991-11e9-968b-0d8b4eb8dcd9.jpg)

LettError:
Thank you for posting this, @justvanrossum GitHub had the hiccups when I tried earlier.

typemytype:
What happens inside: `text(..)` draws the text actually in a `textBox(..)` with a width/height its gets from `textSize`.

Somehow the frame setter thinks the text does not fit inside that box.... and drops the last character.

I assume there is a text wrap setting that can be set while drawBot is setting text with `text(..)`

justvanrossum:
Erik, can you verify whether BezierPath().text() shows the same problem?

LettError:
I was going to and then got lost trying to set feature variation values for BezierPath.

LettError:
Same result when drawing text with bezierPath().text().
 * Blue background: drawing text directly.
 * Green background: drawing with bezierPath().

```python

weightMin = listFontVariations('MutatorMathTest')['wght']['minValue']
weightMax = listFontVariations('MutatorMathTest')['wght']['maxValue']

widthMin = listFontVariations('MutatorMathTest')['wdth']['minValue']
widthMax = listFontVariations('MutatorMathTest')['wdth']['maxValue']

txt = '→→→→'

thisFontName = "MutatorMathTest"
thisSize = 200
[...]

PR #319, opened 2019-10-29
improve usage of text with a given complex formattedString

typemytype:
still todo: updating tests

this should fix a lot of related issues: #315 #254


drawBot/context/baseContext.py
```diff
@@ -886,6 +875,58 @@ def copy(self):
...
+    w += extraPadding
+
+    if align is not None:
+        attributedString = attributedString.mutableCopy()
+        # overwrite all align settings in each paragraph style
+        def block(value, rng, stop):
+            value = value.mutableCopy()
+            value.setAlignment_(FormattedString._textAlignMap[align])
+            attributedString.addAttribute_value_range_(AppKit.NSParagraphStyleAttributeName, value, rng)
+        attributedString.enumerateAttribute_inRange_options_usingBlock_(AppKit.NSParagraphStyleAttributeName, (0, len(attributedString)), 0, block)
```
justvanrossum: Wouldn't `attributedString.addAttribute_value_range_(AppKit.NSParagraphStyleAttributeName, value, rng)` over the whole range without enumerating be working, too?

drawBot/context/baseContext.py
```diff
@@ -886,6 +875,58 @@ def copy(self):
...
+    extraPadding = 20
+    x, y = xy
+    w, h = attributedString.size()
+    w += extraPadding
+
+    if align is not None:
+        attributedString = attributedString.mutableCopy()
+        # overwrite all align settings in each paragraph style
+        def block(value, rng, stop):
+            value = value.mutableCopy()
```
justvanrossum: Why this copy?

drawBot/context/baseContext.py
typemytype: no, cause a NSParagraphStyle can contain different kinds of settings. It could be that  a formattedString contains different line heights... so there are multiple paragraph styles

drawBot/context/baseContext.py
typemytype: this should not modify the original attributed string (which is also the source of the formattedString) 
The copy on the paragraphStyle is also necessary as this seems to not good to modify existing objects inside a attributed string and the copy on the attributed string seems not to be a deep copy where internal objects are also copied.

drawBot/context/baseContext.py
justvanrossum: Ahh, duh, I missed the paragraph style object, got it now. Alignment is not a single simple attr.

drawBot/context/baseContext.py
```diff
@@ -886,6 +874,62 @@ def copy(self):
...
+
+    for i, (originX, originY) in enumerate(origins):
+        ctLine = ctLines[i]
+        rng = CoreText.CTLineGetStringRange(ctLine)
+
+        attributedSubstring = attributedString.attributedSubstringFromRange_(rng)
+        # strip trailing returns
+        if attributedSubstring.string()[-1] in ["\n", "\r"]:
+            rng.length -= 1
+            attributedSubstring = attributedString.attributedSubstringFromRange_(rng)
```
justvanrossum: How about something like:
`attributedString.deleteCharactersInRange_((rng.length - 1, 1))`
It's a mutable attr string, and I would _assume_ that `attributedSubstringFromRange_()` earlier gave you a copy. (I will verify that.)

justvanrossum: Yup, all good:
```python
s = "abcdefghijklmnop"
fs = FormattedString()
fs.append(s)
ats = fs.getNSObject()
assert ats.string() == s
sub = ats.attributedSubstringFromRange_((3, 3))
assert sub.string() == "def"
sub.deleteCharactersInRange_((1, 1))
assert sub.string() == "df"
assert ats.string() == s
```

drawBot/context/baseContext.py
```diff
@@ -886,6 +874,62 @@ def copy(self):
...
+
+    setter = CoreText.CTFramesetterCreateWithAttributedString(attributedString)
+    path = Quartz.CGPathCreateMutable()
+    Quartz.CGPathAddRect(path, None, Quartz.CGRectMake(x, y, w, h * 2))
+    frame = CoreText.CTFramesetterCreateFrame(setter, (0, 0), path, None)
+    ctLines = CoreText.CTFrameGetLines(frame)
+    origins = CoreText.CTFrameGetLineOrigins(frame, (0, len(ctLines)), None)
+    boxes = []
+
+    for i, (originX, originY) in enumerate(origins):
```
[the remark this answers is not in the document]
typemytype: changed

drawBot/context/baseContext.py
typemytype: changed

justvanrossum:
Looks good to me!

justvanrossum:
We should add some tests before merging, though. 

Btw. there's a version number bump here that probably shouldn't go into this PR.

typemytype:
this PR also fixes the ACaslon and https problems listed in #314

PR #320, opened 2019-10-30
CI with github actions!!!!

typemytype:


typemytype:
on fail add a drawbot script that displays the differences between the images and load it up as an asset

typemytype:
github actions works!!! 
on failure all test assets are uploaded and a difference pdf is build

on travis the latest os cannot be picked... so the tests are not correct or comparable
on failure all test assets are ftp-ed to http://static.typemytype.com/drawBot/CItempTestData/

typemytype:
[github action runs on 10.15](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources) and there are no other macOS's to choose from

typemytype:
this PR fixes https://github.com/typemytype/drawbot/issues/358

PR #321, opened 2019-10-31
Add PIL/pillow as a bundled package

justvanrossum:
...pretty please? :)


typemytype:
It will just make it bigger...

PR #325, opened 2019-11-02
Fix traceback when typing enter or backspace in read-only text view

justvanrossum:


typemytype:
damn I was to fast!

the `selectedRange = self.selectedRange()` needs to be called before the super inserts the line

will fix it in the master

drawBot/ui/codeEditor.py
```diff
@@ -763,6 +765,9 @@ def insertNewline_(self, sender):
                 self.insertText_(leadingSpace)
 
     def deleteBackward_(self, sender):
+        if self.lexer() is None:
+            super(CodeNSTextView, self).deleteBackward_
```
typemytype: oh no, I was really to fast...

typemytype:
fixed here https://github.com/typemytype/drawbot/commit/36103d207ed611f3e9357ca800c5a62c14374d8c

PR #326, opened 2019-11-02
Make sure we don't try to use QTKit on 10.15 and up

### LINKED ISSUE #312: QTKit not available anymore on 10.15
```
/drawBot/context/movContext.py", line 2, in <module>
    import QTKit
ModuleNotFoundError: No module named 'QTKit'
```

I would even opt to deprecate mov totally

justvanrossum:
...and give an understandable error message. This should fix #312.

### DIFF
--- a/drawBot/context/movContext.py
+++ b/drawBot/context/movContext.py
@@ -1,10 +1,19 @@
 import AppKit
-import QTKit
 import Quartz
+from distutils.version import StrictVersion
+import platform
+
+osVersionCurrent = StrictVersion(platform.mac_ver()[0])
+if osVersionCurrent >= StrictVersion("10.15"):
+    # QTKit is being deprecated
+    QTKit = None
+else:
+    import QTKit
 
 import os
 
 from drawBot.misc import DrawBotError, warnings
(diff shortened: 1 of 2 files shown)


drawBot/context/movContext.py
```diff
@@ -1,10 +1,19 @@
...
+if osVersionCurrent >= StrictVersion("10.15"):
+    # QTKit is being deprecated
+    QTKit = None
+else:
+    import QTKit
 
 import os
 
 from drawBot.misc import DrawBotError, warnings
+from drawBot.scriptTools import osVersionCurrent
```
typemytype: move the import up and use it to compare the os version

typemytype:
LGTM!

justvanrossum:
No, my scriptTools import was a mistake, a leftover from an earlier version. I reverted to my version (with an order closer to your version), without the scriptTools import. I think it's wrong for a context to import scriptTools, as it's potentially a circular reference.

If we want to centralize os version checking, we should make a small dedicated module at the drawBot top level, say `drawBot/osVersion.py`. But that's for another time.

PR #327, opened 2019-11-03
Centralize getting os version as drawBot.macOSVersion.macOSVersion

justvanrossum:
- Adapt other code to it
- Fixed a wrong version comparison in pdfContext
- Use relative imports here and there

### DIFF
--- a/drawBot/macOSVersion.py
+++ b/drawBot/macOSVersion.py
@@ -3,4 +3,4 @@ import platform
 
 # It is safe to compare osVersion to strings, as StrictVersion casts strings
 # to StrictVersion instances upon compare.
-macOSVersion = StrictVersion(platform.mac_ver()[0])
+macOSVersion = StrictVersion(platform.mac_ver("0.0.0")[0])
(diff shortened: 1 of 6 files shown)


drawBot/macOSVersion.py
```diff
@@ -0,0 +1,6 @@
+from distutils.version import StrictVersion
+import platform
+
+# It is safe to compare osVersion to strings, as StrictVersion casts strings
+# to StrictVersion instances upon compare.
+macOSVersion = StrictVersion(platform.mac_ver()[0])
```
typemytype: Check if the platform is not None and use a fallback instead

See `AttributeError: 'StrictVersion' object has no attribute 'version'` on read the docs

justvanrossum: Yes, I think I found a solution.

justvanrossum: Does RTD build PR branches? If not, we may have to merge, and continue on master if it still doesn't work.

justvanrossum: Ah right, it builds "stable" and "latest". I'm fairly confident this PR will fix the RTD build, though, so I think we should merge (once Travis is done :) ).

PR #328, opened 2019-11-03
Fix for issue 258: text stroke should not scale with font size

### LINKED ISSUE #258: Stroke thickness of text is different from paths
```python
stroke(0.5)
strokeWidth(5)
fill(None)
fontSize(300)
text("Hallo", (100, 100))

bez = BezierPath()
bez.text("Hallo", offset=(100, 400), fontSize=300)
drawPath(bez)
```
image

justvanrossum:
This will fix #258, but is a breaking change.


typemytype:
a breaking change is fine, its was wrong before...  the drawBot.strokeWidth is independent of the font size

PR #338, opened 2019-11-14
Work around non-mutable attributed string

### LINKED ISSUE #337: Problem drawing FormattedString
With this code:

``` python
fs = FormattedString("A\n")
text(fs, (0, 0))
```

I get a traceback on this line:

https://github.com/typemytype/drawbot/blob/04096d968b9522702fa7010096e49206c140ff36/drawBot/context/baseContext.py#L907

```
  File "drawBot/drawBotDrawingTools.pyc", line 1607, in text
  File "drawBot/context/baseContext.pyc", line 906, in makeTextBoxes
AttributeError: 'NSConcreteAttributedString' object has no attribute 'deleteCharactersInRange_'
```

It looks like the FormattedString should be a NSMutableAttributedString but instead it's a NSConcreteAttributedString? However I can't tell how this might have happened.

justvanrossum:
Fixes #337.

### DIFF
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -904,7 +904,8 @@ def makeTextBoxes(attributedString, xy, align, plainText):
         attributedSubstring = attributedString.attributedSubstringFromRange_(rng)
         # strip trailing returns
         if attributedSubstring.string()[-1] in ["\n", "\r"]:
-            attributedSubstring.deleteCharactersInRange_((rng.length - 1, 1))
+            rng.length -= 1
+            attributedSubstring = attributedString.attributedSubstringFromRange_(rng)
         width, height = attributedSubstring.size()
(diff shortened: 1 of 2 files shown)


drawBot/context/baseContext.py
```diff
@@ -904,7 +904,8 @@ def block(value, rng, stop):
         attributedSubstring = attributedString.attributedSubstringFromRange_(rng)
         # strip trailing returns
         if attributedSubstring.string()[-1] in ["\n", "\r"]:
-            attributedSubstring.deleteCharactersInRange_((rng.length - 1, 1))
+            rng.length -= 1
```
typemytype: an other option is to call `attributedSubstring.mutableCopy()` before `deleteCharactersInRange_`

justvanrossum: Same difference, pretty much...

PR #339, opened 2019-11-14
don't add a box when the string is empty; fixes #337 further

### LINKED ISSUE #337: Problem drawing FormattedString

justvanrossum:


justvanrossum:
This may do the wrong thing for the line distance in the empty line, so perhaps indeed the `\n` shoult be stripped later.

justvanrossum:
Yeah, this is definitely not correct wrt. the line height of the empty line. Not immediately seeing the right solution, though.

A line distance test case should cover:
```python
fs = FormattedString("A\n\nC")
text(fs, (0, 0))
```
vs.

```python
fs = FormattedString("A\nB\nC")
text(fs, (0, 0))
```

typemytype:
LGTM!

PR #343, opened 2019-12-06
show the error message when it failes

### LINKED ISSUE #342: Updater may be lying sometimes
Whenever the current version number cannot be retrieved for some reason, there will be a message proudly proclaiming "You have the latest version!", which is then very unlikely to be the truth... See #341.

typemytype:
fixes #342

### DIFF
--- a/drawBot/updater.py
+++ b/drawBot/updater.py
@@ -18,14 +18,16 @@ from .misc import DrawBotError, getDefault
 
 _versionRE = re.compile(r'__version__\s*=\s*\"([^\"]+)\"')
 
+__fallback_version__ = "0.0"
+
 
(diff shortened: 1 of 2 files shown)


drawBot/updater.py
```diff
@@ -36,14 +38,17 @@ def getCurrentVersion():
         # in py3 this are bytes and a string object is needed
         code = str(code.decode("ascii"))
         response.close()
-    except Exception:
+    except Exception as e:
         # just silently fail, its not so important
+        errors.append(f"Cannot retrieve version number from DrawBot repo:\n{e}")
```
justvanrossum: The preceding comment needs adjusting...

PR #346, opened 2019-12-12
Check that a path has elements before optimizing

andyclymer:
A situation can arise where a BezierPath has no elements, resulting in a traceback in the optimizePath function:

```python
import drawBot as db
db.newDrawing()
path = db.BezierPath()
path.textBox("", (0, 0, 100, 100))
db.drawPath(path)
```

```
  File "<untitled>", line 4, in <module>
  File "/Users/clymer/Documents/Code/Git repos/GitHub/typemytype/drawbot/drawBot/context/baseContext.py", line 361, in textBox
  File "/Users/clymer/Documents/Code/Git repos/GitHub/typemytype/drawbot/drawBot/context/baseContext.py", line 452, in optimizePath
IndexError: NSRangeException - elementAtIndex:associatedPoints:: index (-1) beyond bounds (0)
```

This PR optimizes the path only if the element count is greater than zero.

### DIFF
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -449,21 +449,21 @@ class BezierPath(BasePen):
 
     def optimizePath(self):
         count = self._path.elementCount()
-        if count > 0:
-            if self._path.elementAtIndex_(count - 1) == AppKit.NSMoveToBezierPathElement:
-                optimizedPath = AppKit.NSBezierPath.alloc().init()
-                for i in range(count - 1):
-                    instruction, points = self._path.elementAtIndex_associatedPoints_(i)
-                    if instruction == AppKit.NSMoveToBezierPathElement:
-                        optimizedPath.moveToPoint_(*points)
-                    elif instruction == AppKit.NSLineToBezierPathElement:
-                        optimizedPath.lineToPoint_(*points)
-                    elif instruction == AppKit.NSCurveToBezierPathElement:
-                        p1, p2, p3 = points
-                        optimizedPath.curveToPoint_controlPoint1_controlPoint2_(p3, p1, p2)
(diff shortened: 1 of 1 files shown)


drawBot/context/baseContext.py
```diff
@@ -449,20 +449,21 @@ def controlPointBounds(self):
...
-                    optimizedPath.moveToPoint_(*points)
-                elif instruction == AppKit.NSLineToBezierPathElement:
-                    optimizedPath.lineToPoint_(*points)
-                elif instruction == AppKit.NSCurveToBezierPathElement:
-                    p1, p2, p3 = points
-                    optimizedPath.curveToPoint_controlPoint1_controlPoint2_(p3, p1, p2)
-                elif instruction == AppKit.NSClosePathBezierPathElement:
-                    optimizedPath.closePath()
-            self._path = optimizedPath
+        if count > 0:
```
justvanrossum: I would prefer this:
```python
    count = self._path.elementCount()
    if not count:
        return
    ...
```

Or `count == 0` if you prefer that. Bailing out early reduces the indentation level and is a little clearer to me.

drawBot/context/baseContext.py
```diff
@@ -449,20 +449,21 @@ def controlPointBounds(self):
...
+            if self._path.elementAtIndex_(count - 1) == AppKit.NSMoveToBezierPathElement:
```
justvanrossum: Or maybe this:
```python
    if not count or self._path.elementAtIndex_(count - 1) != AppKit.NSMoveToBezierPathElement:
        return
```

andyclymer: I pushed an update like this

justvanrossum:
Thanks Andy!

typemytype:
thanks!!!

PR #348, opened 2019-12-20
support for viewbox

### LINKED ISSUE #347: viewBox for svg
see https://forum.drawbot.com/topic/225/svg-attributes-viewbox

proposal: ˋsaveImage(svgPath, viewBox=(x, y, w, h))ˋ

as and extra arg while saving a svg, the viewBox can be different from the width and height of the document.

typemytype:
fixes #347


drawBot/context/svgContext.py
```diff
@@ -316,7 +316,7 @@ def _newPage(self, width, height):
         self._svgContext = XMLWriter(self._svgData, encoding="utf-8", indentwhite=self.indentation)
         self._svgContext.width = self.width
         self._svgContext.height = self.height
-        self._svgContext.begintag("svg", width=self.width, height=self.height, **self._svgTagArguments)
+        self._svgContext.begintag("svg", width=self.width, height=self.height, viewbox=f"0 0 {self.width} {self.height}", **self._svgTagArguments)
```
justvanrossum: Suggest to specify the attrs as a list of (attr, value) tags, so the order becomes controllable. It looks rather odd that `width` and `height` aren't together and in that order in the output.

For example:

```python
        attrs = [('width', self.width), ('height', self.height), ('viewbox', f"0 0 {self.width} {self.height}")]
        self._svgContext.begintag("svg", attrs + self._svgTagArguments)
```
`self._svgTagArguments` would also need to become a list of tuples.

typemytype: oh, that is the way to specify the order, always thought it was alphabetically

justvanrossum: Btw. it's `viewBox`, not `viewbox`.

PR #350, opened 2019-12-21
points in a bezierPath must be a tuple as it is immutable

typemytype:
+ improve docs

### DIFF
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -82,9 +82,9 @@ class BezierContour(list):
             pen.closePath()
 
     def _get_points(self):
-        return [point for segment in self for point in segment]
+        return tuple([point for segment in self for point in segment])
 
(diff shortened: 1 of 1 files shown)


drawBot/context/baseContext.py
```diff
@@ -699,22 +699,22 @@ def _points(self, onCurve=True, offCurve=True):
...
                 pts = pts[-1:]
             points.extend([(p.x, p.y) for p in pts])
-        return points
+        return tuple(points)
 
     def _get_points(self):
         return self._points()
 
-    points = property(_get_points, doc="Return a list of all points.")
+    points = property(_get_points, doc="Return a immutable list of all points in the bezierPath as point coordinate `(x, y)` tuples.")
```
justvanrossum: `a` -> `an`, same in the three following cases.

justvanrossum: Also: `BezierPath` (cap)

PR #357, opened 2020-01-28
add svgID svgClass svgLink to BezierPath and FormattedString objects

typemytype:
todo:

- [x] test
- [x] see how it ends up in the docs

### DIFF
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -475,7 +475,9 @@ class BezierPath(BasePen):
         return new
 
     def _copyContextProperties(self, other):
-        pass
+        self.svgID = other.svgID
+        self.svgClass = other.svgClass
+        self.svgLink = other.svgLink
 
     def reverse(self):
         """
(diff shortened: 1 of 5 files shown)


typemytype:
fixes #357

drawBot/context/baseContext.py
```diff
@@ -747,6 +753,32 @@ def __iter__(self):
...
 
+    # context specific attributes
+
+    def _get_svgID(self):
+        return getattr(self, "_svgID", None)
+
+    def _set_svgID(self, value):
+        self._svgID = value
+
+    svgID = property(_get_svgID, _set_svgID, doc="The path svg id, as a string.")
```
justvanrossum: I would get rid of the word "path" in the doc string, and use a mixin for these properties so you can share the code between FormattedString and BezierPath. Right now you also write "path" in the FormattedString version.

Or use something like this to reduce boilerplate:

```python
class docprop:

    def __init__(self, doc):
        self.__doc__ = doc

    def __set_name__(self, owner, name):
        self.name = name

    def __get__(self, obj, cls=None):
        return obj.__dict__.get(self.name)

    def __set__(self, obj, value):
        obj.__dict__[self.name] = value

    def __delete__(self, obj):
        obj.__dict__.pop(self.name, None)


class Foo:
    svgID = docprop("this is the svg id")

f = Foo()
print(f.svgID)
f.svgID = "lalala"
print(f.svgID)
del f.svgID
print(f.svgID)
```

typemytype:
those new attributes are popping up in the docs!!!

Issue #375, opened 2020-04-17
support for antiAliasing while saveImage


typemytype:
a proposal

`saveImage(path, antiAliasing=False)`

discussion started on forum: https://forum.drawbot.com/topic/251/turn-off-antialiasing/3

![Screen Shot 2020-04-17 at 15 40 02](https://user-images.githubusercontent.com/1190358/79575503-30441380-80c2-11ea-9bd9-af77fd74aa42.png)

```python
size(100, 100)
#fill(1)
#rect(0, 0, 100, 100)
fill(0)
oval(10, 10, 40, 80)
stroke(0)
line((-0.5, -0.5), (100.5, 100.5))
line((0, 20.5), (100, 20.5))

fontSize(10)
text("a", (62, 30))

saveImage('testAliasingNormal.png')

pdf = pdfImage()
page = pdf.pageAtIndex_(0)

import AppKit, Quartz

def _makeBitmapImageRep(image, page=None, imageResolution=72.0, colorSpaceName=AppKit.NSCalibratedRGBColorSpace, antiAliasing=True):
    """Construct a bitmap image representation at a given resolution."""
    scaleFactor = max(1.0, imageResolution) / 72.0
    rep = AppKit.NSBitmapImageRep.alloc().initWithBitmapDataPlanes_pixelsWide_pixelsHigh_bitsPerSample_samplesPerPixel_hasAlpha_isPlanar_colorSpaceName_bytesPerRow_bitsPerPixel_(
        None,                                    # planes
        int(image.size().width * scaleFactor),   # pixelsWide
        int(image.size().height * scaleFactor),  # pixelsHigh
        8,                                       # bitsPerSample
        4,                                       # samplesPerPixel
        True,                                    # hasAlpha
        False,                                   # isPlanar
        colorSpaceName,                          # colorSpaceName
        0,                                       # bytesPerRow
        0                                        # bitsPerPixel
    )
    rep.setSize_(image.size())

    AppKit.NSGraphicsContext.saveGraphicsState()
    try:
        AppKit.NSGraphicsContext.setCurrentContext_(
            AppKit.NSGraphicsContext.graphicsContextWithBitmapImageRep_(rep))
        context = AppKit.NSGraphicsContext.currentContext().CGContext()
        if not antiAliasing:
            Quartz.CGContextSetInterpolationQuality(context, Quartz.kCGInterpolationNone)
            Quartz.CGContextSetAllowsAntialiasing(context, False)
        if page is not None:
[...]

PR #376, opened 2020-04-17
Anti aliasing

typemytype:


typemytype:
fix #376

typemytype:
fix #375

typemytype:
changelog text:

```
- Add support for `antiAliasing` while calling `saveImage` when exporting to an image.
```

PR #379, opened 2020-04-27
add paragraph spacing to makeTextBoxes

typemytype:
needs some specific test


justvanrossum:
Why did the test results change? The line distance is now different. You can't just change the expected results to match the new implementation...

typemytype:
the previous version was wrong

this is clear when a textBox is drawn next to it:

```python
import drawBot
drawBot.newDrawing()
fs = drawBot.FormattedString("A\nB\nC\n")
drawBot.text(fs, (0, 60))
fs = drawBot.FormattedString("A\n\nC\n")
drawBot.text(fs, (10, 60))
drawBot.textBox(fs, (20, 10, 60, 60))
drawBot.endDrawing()
```

The "C" jumps in the current version. Drawing with `text(..)` should have the same result hard returns line breaks as drawing with `textBox(...)` 

Screen Shot 2020-05-31 at 12 51 31 
Screen Shot 2020-05-31 at 12 50 28

PR #384, opened 2020-05-13
add support for isContinuous

### LINKED ISSUE #383: improve support for "Apply/Build/Update" button in Variable
add an option to switch off the callback for ui elements, proposal add `isContinuously` (default to 
`True` to keep the current behaviour.

```python
Variable([
    dict(name="foo", ui="Slider", isContinuously=False),
    dict(name="Update", ui="Button")
], globals())
```

typemytype:
option to disable a direct callback of sliders and other control changes

fixes #383

### DIFF
--- a/drawBot/drawBotDrawingTools.py
+++ b/drawBot/drawBotDrawingTools.py
@@ -2377,7 +2377,7 @@ class DrawBotDrawingTool(object):
         as you want to insert the variable in the current workspace.
         It is required that `workSpace` is a `dict` object.
 
-        The `continuous` argument controls whether script is run when UI
+        The `continuous` argument controls whether the script is run when UI
         elements change. The default is `True`, which will execute the script
         immediately and continuously when the user input changes. When set to
         `False`, there will be an "Update" button added at the bottom of the window.
--- a/drawBot/drawBotDrawingTools.py
+++ b/drawBot/drawBotDrawingTools.py
@@ -2377,7 +2377,7 @@ class DrawBotDrawingTool(object):
         as you want to insert the variable in the current workspace.
         It is required that `workSpace` is a `dict` object.
 
-        The `continuous` argument controls whether whether script is run when UI
+        The `continuous` argument controls whether script is run when UI
         elements change. The default is `True`, which will execute the script
         immediately and continuously when the user input changes. When set to
         `False`, there will be an "Update" button added at the bottom of the window.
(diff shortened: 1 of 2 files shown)


typemytype:
test are fixed in different PR...

gferreira:
⚠️ `isContinuous`

drawBot/drawBotDrawingTools.py
```diff
@@ -2369,14 +2369,18 @@ def ImageObject(self, path=None):
...
         """
         Build small UI for variables in a script.
 
         The `workSpace` is usually `globals()`
         as you want to insert the variable in the current workspace.
         It is required that `workSpace` is a `dict` object.
 
+        Setting `isContinuous` to `False` will add a "Update" button at the bottom.
+        for the user to click to get and update. The default is `True` and will
+        set all control updating the script continuously.
```
justvanrossum, line 2382: This needs editing. Perhaps:
"""
The `isContinuous` argument controls whether whether script is run when UI elements change. The default is `True`, which will execute the script immediately and continuously when the user input changes. When set to `False`, there will be an "Update" button added at the bottom of the window. The user will have to click this button to execute the script and see the changes. This is useful when the script is slow, and continuous execution would decrease responsiveness.
"""

justvanrossum: Btw. I think I would still prefer to call it `continuous`, to be more consistent with Vanilla.

typemytype: changed!

drawBot/drawBotDrawingTools.py
```diff
@@ -2369,14 +2369,22 @@ def ImageObject(self, path=None):
...
-    def Variable(self, variables, workSpace):
+    def Variable(self, variables, workSpace, continuous=True):
         """
         Build small UI for variables in a script.
 
         The `workSpace` is usually `globals()`
         as you want to insert the variable in the current workspace.
         It is required that `workSpace` is a `dict` object.
 
+        The `continuous` argument controls whether whether script is run when UI
```
justvanrossum, line 2380: "whether whether script" -> "whether the script"

drawBot/drawBotDrawingTools.py
```diff
@@ -2369,14 +2369,22 @@ def ImageObject(self, path=None):
...
-    def Variable(self, variables, workSpace):
+    def Variable(self, variables, workSpace, continuous=True):
         """
         Build small UI for variables in a script.
 
         The `workSpace` is usually `globals()`
         as you want to insert the variable in the current workspace.
         It is required that `workSpace` is a `dict` object.
 
+        The `continuous` argument controls whether script is run when UI
```
justvanrossum, line 2380: "whether script is" -> "whether the script is"

PR #392, opened 2020-06-13
improve support by launching the app with corefoundationasyncio

typemytype:


typemytype:
more info https://github.com/alberthier/corefoundationasyncio

requirements.txt
```diff
@@ -1,5 +1,6 @@
 setuptools>=45.2.0
 pyobjc
+CoreFoundationEventLoop
```
justvanrossum, line 3: This should be `corefoundationasyncio`

typemytype: oh offcourse...

colinmford:
so psyched about this

justvanrossum:
> so psyched about this

It's not a super easy paradigm to get into, but worth learning. FontGoggles uses it throughout to keep the UI responsive, if you'd like some non-trivial example code.

PR #393, opened 2020-06-14
draw a white background rect when exporting to mp4

### LINKED ISSUE #391: default mp4 background color
👋

if I run this code in drawbot

```python
radius = 125
for angle in range(360):
    newPage(400, 400)
    translate(width()/2, height()/2)
    xx = cos(radians(angle)) * radius
    yy = sin(radians(angle)) * radius
    oval(xx-3, yy-3, 6, 6)
saveImage('ball.mp4')
```

there is a difference between the mp4 background (black) and the canvases backgrounds in the app (white).
I imagine it happens because the pdf canvases are transparent, not white, and transparent backgrounds are converted in black by ffmpeg. But that's counter intuitive for beginners, don't you think? If anyone exports a video with the default drawbot settings there is no visible output.

typemytype:
fixes #391

### DIFF
--- a/drawBot/context/mp4Context.py
+++ b/drawBot/context/mp4Context.py
@@ -31,6 +31,9 @@ class MP4Context(PNGContext):
     def _newPage(self, width, height):
         super(MP4Context, self)._newPage(width, height)
         self._frameDurations.append(self._defaultFrameDuration)
+        # https://github.com/typemytype/drawbot/issues/391
+        # draw a solid white background by default
+        # ffmpeg converts transparency to a black background color
         self.save()
         self.fill(1)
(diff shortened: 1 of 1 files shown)


drawBot/context/mp4Context.py
```diff
@@ -31,6 +31,10 @@ def _frameDuration(self, frameDuration):
     def _newPage(self, width, height):
         super(MP4Context, self)._newPage(width, height)
         self._frameDurations.append(self._defaultFrameDuration)
+        self.save()
```
justvanrossum, line 34: Suggest to add a comment why we do this.

PR #415, opened 2020-12-17
Update drawBot.jpg URL

shakeel:
The current cloudfront.net drawBot.jpg URL is broken, replaced it with the raw asset URL from this github repo.


typemytype:
ha nice!

could you also change other references to this file, this is making the tests fail:

like in this test script: https://github.com/typemytype/drawbot/blob/master/tests/drawBotScripts/imageHTTP.py

thanks!

typemytype:
ping! 

could you also change the url in other text scripts?

thanks

Issue #112, opened 2017-08-25
Fonts are not reloaded even when changing on disk


frankrolf:
I am using DrawBot to create proofs, the fonts used in those proofs are re-generated frequently.
When I run the script from the DrawBot window, the font file that is picked up first stays active forever.

My script is wrapped like this:
```
newDrawing()
	for fontFile in fontFiles:
		newPage()
		fs = FormattedString(
			proofText, fontFile
		)
		text(fs)
	for fontFile in fontFiles:
		uninstallFont(fontFile)
endDrawing()
```
I explicitly added the `newDrawing()` and `endDrawing()` hoping it would have some kind of effect, but no.
It works as expected when I run the script from the command line, but it sometimes is nice to have that visual output …

typemytype:
I guess this is an apple CoreText issue... it seems like the fonts are cached. When installing fonts from the same source (path on disk) with the same `fontName` (aka postscriptName) within the same runtime of an app, their is no visual update... which is idd not nice...

justvanrossum:
How does RF manage to get this right with test install? Seems like it should be more or less the same mechanism.

frankrolf:
Here is a tangentially related problem:
I want to compare revisions of the same design, and pick up two OTF files that have different file names but same PS name. Only the first one gets picked up throughout the whole output PDF (and GIF, etc). When I edit the PS name, it works as expected.
I assume this is a limitation within the PDF format? Any suggestions on how to modify the PS name on the fly?

typemytype:
I guess you can embed two fonts with the same psname... 

* open with fontTools
* edit the name table
* store it in a separate folder
* install from from
* remove that folder

frankrolf:
That works so well! Thank you Frederik!

typoman:
See:
https://github.com/typemytype/drawbot/issues/261#issuecomment-463424549
I have similar issue with the app. But I'm going to see if I can reprodue it using the module too.

typoman:
I have tested drawbot as a module and there is no font caching happening in my case when I generate a font and change it constantly and use it in a text inside drawbot. I also tested the script simultaneously inside the app and it seems font caching is persistent there. I don't know if this helps but there is a way to clear the font cache in the shell but maybe it's not what drawBot app is supposed to do or has the privilege to do?

Clear font cache in shell:
```
atsutil databases -removeUser && \
atsutil server -shutdown && \
atsutil server -ping
```

typemytype:
[...]

PR #421, opened 2021-03-21
Fix for #261: don't try to install fonts from paths, but build NSCTFont directly from file

### LINKED ISSUE #261: Ad-hoc installation of font works every 2nd time
```
my_text = FormattedString(
    'I love DrawBot',
    font='my_font.otf')
    
text(my_text, (100, 100))
```

This works the first time, then fails with
```*** DrawBot warning: font: 'my_font' is not installed, back to the fallback font: 'LucidaGrande' ***```
Then works again, then fails … etc.

justvanrossum:
- When using font paths, don't try to install, but make NSCTFont object directly from file
- Add support for ttc/otc/dfont for font paths (this adds fontNumber properties/arguments in various places)
- Some refactoring with the font properties (such as fontDescender) to reduce code
- Delete skipped fontPath.svg test file

This should fix #261


justvanrossum:
I was hoping this would also fix #112, but it doesn't :(

typemytype:
ooh cool!

I would deprecate these callbacks: https://github.com/typemytype/drawbot/blob/improve-ttc-and-fix-261/drawBot/drawBotDrawingTools.py#L2215

maybe deprecate hard: they dont need to work anymore
this will also remove all the `_tempInstalledFonts` references in the drawingTool class 

about fixing #112: There is no caching anymore in DrawBot and by the OS, so different fonts with the same postscript name should appear different from each other... strange

justvanrossum:
People may be using `installFont()`, so I would prefer to leave it for a while and emit a deprecation warning.

> about fixing #112: There is no caching anymore in DrawBot and by the OS, so different fonts with the same postscript name should appear different from each other... strange

Well apparently there is, so the "by the OS" is not true (the DB cache gets reset after each drawing). I have an idea I want to try, but that will be for another PR (if it works at all).

justvanrossum:
Ok, I think this is good to go.

Issue #56, opened 2016-07-20
Font caching? 


aaronbell:
I wrote to Erik about this yesterday, but figured I'd post it here too. I'm running into what appears to be a font caching problem.

Script in question (adapted from Erik's sample script):
`
path = PATH_TO_OTF
fontName = installFont(path)
size(1000, 1000)

thisFontSize = 242
s = FormattedString(
    txt="fe AV,",
    font="",
    fontSize=312,
)
textBox(s , (50, 50, width()-50, height()-1.5*50))
uninstallFont(path)
`

To repro:
1) Open any font editor and generate an OTF.
2) Replace the PATH_TO_OTF above with the path to the OTF. 
3) Run the script. 
4) Edit the font, increasing kerning between A and V by 200 and regenerate the OTF. 
5) Run the script. 
6) _Note that nothing has changed in the viewport_
7) Close Drawbot. 
8) Reopen Drawbot. 
9) Run script. 
10) _Note that the kern change performed in step 4 is now visible_

Similarly, instead of closing and reopening drawbot, change the font being used from the non-installed font to something like Helvetica. Run the script once, then change it back. The changes will be visible. 

My guess is that there is some font caching going on and Drawbot does not re-load the font each time the script is run. It could also be that I'm not uninstalling the font properly.

justvanrossum:
Duplicate of #112. Or rather vice versa, but there's more discussion there, so let's continue there.

PR #422, opened 2021-03-24
Reload "path" font when the file changed on disk

### LINKED ISSUE #112: Fonts are not reloaded even when changing on disk

justvanrossum:
This fixes #112.

I managed to do this efficiently without requiring an explicit `reloadFont()` function, byt looking at the modification time at the right moments.

### DIFF
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -2639,13 +2639,20 @@ def getNSFontFromNameOrPath(fontNameOrPath, fontSize, fontNumber):
     return CoreText.CTFontCreateWithFontDescriptor(descriptors[fontNumber], fontSize, None)
 
 
+#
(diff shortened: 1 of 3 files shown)


justvanrossum:
It also fixes #56, which was closed as a duplicate.

PR #425, opened 2021-04-15
adds resetFeatures to openTypeFeatures docs + updates example

roberto-arista:
👋

Hey there!
Yesterday during a class, I noticed that `openTypeFeatures(None)` is deprecated, but `resetFeatures` arg is not documented.  

I just added a sentence to the docstring and I updated the example. I assume the image on the website will update automatically? Otherwise let me know if I need to make other steps.

What do you think?


roberto-arista:
🎉

Issue #429, opened 2021-07-22
Font path error after upgrading to 3.127 


gorjious:
Upgraded to 3.127 from 3.126 and now getting this error in a file that was previously working:
`TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType`

Is this a bug or any suggestions on how to fix this?

The full error message is this:
```
*** DrawBot warning: font: 'MyFont-Regular' can't be found, using the fallback font 'LucidaGrande' ***
Traceback (most recent call last):
  File "drawbot_test.py", line 352, in <module>
  File "drawBot/drawBotDrawingTools.pyc", line 1893, in FormattedString
  File "drawBot/context/baseContext.pyc", line 1091, in __init__
  File "drawBot/context/baseContext.pyc", line 1185, in append
  File "drawBot/context/baseContext.pyc", line 1105, in _setAttribute
  File "drawBot/context/baseContext.pyc", line 1433, in font
  File "drawBot/misc.pyc", line 374, in wrapper
  File "drawBot/context/baseContext.pyc", line 2630, in getNSFontFromNameOrPath
  File "genericpath.pyc", line 19, in exists
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

justvanrossum:
Can you post a complete script that reproduces this? It may be specific to the font. If so: can you share the font?

gorjious:
I'm thinking it was an issue with absolute vs relative paths. I notice that leaving out the first slash causes this error so that may have been it. 

However, I am trying to make a simple example to reproduce it, and am running into an issue that in 3.127 "font can't be found" and it doesn't render but in 3.126 it says "font is not installed" yet it still renders. 

<img width="1280" alt="Screenshot 2021-07-22 at 16 59 29" src="https://user-images.githubusercontent.com/23149420/126643324-de0d80c3-b60e-41b4-b680-cab9d4b662ad.png">

What should be the correct way of linking both absolute and relative paths?

[drawbot_test.zip](https://github.com/typemytype/drawbot/files/6862367/drawbot_test.zip)

justvanrossum:
Ah, I think I see what's going on: you should not use the _output_ of `font()` as another _input_ for `font()`. Just use the full path throughout.

Relative file names are relative to the script's parent folder. So if you use "document_fonts/myfont.otf" there should be a folder called "document_fonts" in the same folder as where the script is saved.

[...]

PR #430, opened 2021-07-22
Improve error messages for ill-specified fonts

justvanrossum:
This should improve error messages for #429:
- warn if a font can't be located by name nor by path
- raise TypeError if `fontNameOrPath` argument is neither a string nor a path-like object


justvanrossum:
The failings are unrelated to this PR, but I can't reproduce them locally, even though I run the same OS as the GHA bot: 10.15.7. Will need to investigate :( (Or disable the three failing tests)

justvanrossum:
Test failings were fixed by #431

Issue #402, opened 2020-09-29
Appending a glyph to a FormattedString does not work well when fontVariations are set


MrBrezina:
```
fs = db.FormattedString()
fs.font("Handjet-Regular")
fs.fontSize(350)
fs.fontVariations(wght=300)
fs.appendGlyph("A")
text(fs, (100, 100))
```

Draws a `.notdef` glyph from another font (Lucida?). Everything works as expected if I leave out this line `fs.fontVariations(wght=300)`.

In case it is an issue in the [Handjet font](https://github.com/rosettatype/handjet/blob/master/README.md), let me know, please.

Another related, unexpected behaviour:

I thought I could set up the FormattedString like this:

```
fs = db.FormattedString("", font="Handjet-Regular", fontSize=350, …)
fs.append("xyz")
```

But that does not work, neither with `append()` nor with `appendGlyph()`. The `font` and `fontSize` settings are ignored for the text “xyz“. I would expect them to get carried on.

typemytype:
also read https://forum.drawbot.com/topic/316/appending-a-glyph-to-a-formattedstring-does-not-work-well-when-fontvariations-are-set

an example with Skia

```python
fs = FormattedString()
fs.font("Skia")
fs.fontSize(350)
fs.fontVariations(wght=0.93)
fs += "A"
fs.appendGlyph("A")
text(fs, (100, 100))
```

![image](https://user-images.githubusercontent.com/1190358/128421820-563dcffb-ca89-4849-a29e-7db4611f726a.png)

PR #433, opened 2021-08-06
Set font variations settings while extracting a glyph by name

typemytype:
fixing #402

font variations settings must be set before calling `nsFont.glyphWithName_` to get a nsGlyphInfo with the correct font variation


typemytype:
will also add a test

justvanrossum (approved):

Issue #437, opened 2021-09-24
UI Idea: a "Prettify Source" menu


justvanrossum:
Or how to call it. "Beautify"? "Reformat"?

It should be easy to implement with https://pypi.org/project/black/

justvanrossum:
(This is not my idea, but is a question from a student, and I think it's a good one.)

typemytype:
+1 to make code more readable :)

PEP 8 refers to *Style Guide* so maybe 

* Code/Source Styling
* Clean Up Code/Source
* Optimise Code/Source

justvanrossum:
"Clean up code" is my favorite among those three. I find "Format code" also an option.

PR #439, opened 2021-09-27
Adding support for formatting code

typemytype:
fixing #437


drawBot/ui/drawBotController.py
```diff
@@ -167,6 +167,19 @@ def checkSyntax(self, sender=None):
         self.stdout = None
         self.stderr = None
 
+    def formatCode(self, sender=None):
+        import black
+        # get the code
+        code = self.code()
+        # format the code with black
+        formattedCode = black.format_str(code, mode=black.Mode())
```
justvanrossum, line 175: `black.format_str()` can raise `black.InvalidInput`, for example if the source contains a syntax error.

justvanrossum:
Would it make sense to set the selection back to what it was before, even if we know that can't be precise? Right now it always goes to the end of the script, which feels a bit annoying if the script is long.

drawBot/ui/drawBotController.py
```diff
@@ -167,6 +167,22 @@ def checkSyntax(self, sender=None):
...
+        # get the code
+        code = self.code()
+        # format the code with black
+        try:
+            formattedCode = black.format_str(code, mode=black.Mode())
+        except black.InvalidInput:
+            return
+        # set it back in the text view
+        textView = self.codeView.getNSTextView()
+        # selecte all
```
justvanrossum, line 181:
```suggestion
        # select all
```

justvanrossum:
Shall we merge? This seems to work quite nice.

Issue #451, opened 2021-11-29
proposal saveImage with support for PIL and nsImage objects


typemytype:
started here #363 

proposal to support PIL image objects, and NSImage object

`imObject = saveImage("PIL")` using the the "extension" `PIL` to get a pil context returning a pil image objects

the same for and NSImage:

`imOjbect = saveImage("NSImage")`

as default `saveImage(...)` will return `None`

this means (the deprecated) [support for multiple paths in a single saveImage](https://github.com/typemytype/drawbot/blob/master/drawBot/drawBotDrawingTools.py#L406-L416) will be removed, otherwise its not possible to return an object from a specific context.

PR #452, opened 2021-11-30
Pil nsimage support

typemytype:
add support for

```python
pilImageObject = saveImage("PIL")
nsImageImageObject = saveImage("NSImage")
```

#451


typemytype:
I guess this PR is ready!

drawBot/context/__init__.py
```diff
@@ -17,7 +18,9 @@
     BMPContext,
     MP4Context,
     ICNSContext,
-    PrintContext
+    PrintContext,
+    PILContext,
+    NSImageContext
```
justvanrossum, line 23:
```suggestion
    NSImageContext,
```

drawBot/context/gifContext.py
```diff
@@ -43,7 +44,13 @@ def _writeDataToFile(self, data, path, options):
             options["multipage"] = True
             tempPath = tempfile.mkstemp(suffix=".gif")[1]
 
-        inputPaths = super(GIFContext, self)._writeDataToFile(data, tempPath, options)
+        self._inputPaths = []
+        super(GIFContext, self)._writeDataToFile(data, tempPath, options)
```
justvanrossum, line 48:
```suggestion
        super()._writeDataToFile(data, tempPath, options)
```

drawBot/context/tools/pilTools.py
```diff
@@ -0,0 +1,11 @@
+import AppKit
+import io
+
+
+def pilToNSImage(pilImage):
+    buffer = io.BytesIO()
+    pilImage.save(buffer, pilImage.format)
```
justvanrossum, line 7: Why keep/use `pilImage.format`? What if the original file was in a format that PIL can read, but NSImage cannot? I'd suggest to use PNG here. There's for example no need to recompress to JPG, as NSImage will have to unpack into a pixel array anyway.

drawBot/drawBotDrawingTools.py
```diff
@@ -394,6 +394,8 @@ def saveImage(self, path, *args, **options):
             saveImage("~/Desktop/firstImage300.png", imageResolution=300)
 
         """
+        if not isinstance(path, (str, os.PathLike)):
+            raise TypeError("Cannot apply saveImage options to multiple output formats, expected 'str', got '%s'" % type(path).__name__)
```
justvanrossum, line 398:
```suggestion
            raise TypeError("Cannot apply saveImage options to multiple output formats, expected 'str' or 'os.PathLike', got '%s'" % type(path).__name__)
```

tests/testExport.py
```diff
@@ -210,17 +211,12 @@ def test_saveImage_unknownContext(self):
...
-                drawBot.saveImage(["foo.abcde"])
-        self.assertEqual(output.lines(), ['*** DrawBot warning: saveImage([path, path, ...]) is deprecated, use multiple saveImage statements. ***'])
-        self.assertEqual(cm.exception.args[0], "Could not find a supported context for: 'abcde'")
 
     def test_saveImage_pathList(self):
         self.makeTestDrawing()
         with self.assertRaises(TypeError) as cm:
             drawBot.saveImage(["foo.abcde"], foo=123)
-        self.assertEqual(cm.exception.args[0], 'Cannot apply saveImage options to multiple output formats.')
+        self.assertEqual(cm.exception.args[0], "Cannot apply saveImage options to multiple output formats, expected 'str', got 'list'")
```
justvanrossum, line 219:
```suggestion
        self.assertEqual(cm.exception.args[0], "Cannot apply saveImage options to multiple output formats, expected 'str' or 'os.PathLike', got 'list'")
```

justvanrossum (changes requested):

drawBot/context/gifContext.py
typemytype: the old style `super(class, self)` happens in more places, will change this one, but the rest has to go in a separate PR

#453

drawBot/context/tools/pilTools.py
typemytype: +1 for PNG

typemytype: now internally the image objects are build with from the TIFF representation

typemytype: see https://github.com/typemytype/drawbot/pull/452/files#diff-7c0b8578fa67d5d3f0a2ea4b08902cf57175c06b1b8fd6b3cf0c275d804417faR13

typemytype: an maybe `pilToNSImage`  has to move to `drawBot.misc` which is a bit more accessible then `drawBot.context.tools.pilTools`

drawBot/context/imageObjectContext.py
```diff
@@ -0,0 +1,43 @@
...
+import io
+import AppKit
+
+from .imageContext import ImageContext
+
+class BaseImageObjectContext(ImageContext):
+
+    def _writeDataToFile(self, data, path, options):
+        self._imageObjects = []
+        with tempfile.NamedTemporaryFile(suffix=".tiff") as temp:
```
justvanrossum, line 13: Is there any way we can do thins without a temp file?

typemytype: possible, now nothing is written to disk, let me check

drawBot/context/imageObjectContext.py
```diff
@@ -0,0 +1,42 @@
...
+
+from .imageContext import ImageContext
+
+class BaseImageObjectContext(ImageContext):
+
+    def _writeDataToFile(self, data, path, options):
+        self._imageObjects = []
+        # we just need a path with a file extension
+        # nothing will be written to disk
+        super()._writeDataToFile(data, "temp.tiff", options)
```
justvanrossum, line 13: Are you sure it doesn't now write to a file named "temp.tiff"?

typemytype: separating writing outside the big loop helps to reuse everything from the ImageContext class

justvanrossum: But now there are to methods with "ToFile" in their names which may not actually write "to file". Perhaps something for a later refactoring. I appreciate the code reuse, but it's not very clear what's going on.

typemytype: true, what about removing `toFile`:

_saveImageDataToFile --> _storeImageData, _saveImageData

typemytype: method renamed to _storeImageData

drawBot/context/gifContext.py
```diff
@@ -1,5 +1,6 @@
 import Quartz
-
+import os
+import glob
```
justvanrossum, line 3: `os` and `glob` seem unused.

justvanrossum (approved):

justvanrossum:
Is this one ready to go?

typemytype:
yep go!

Issue #449, opened 2021-11-26
Some drawBot methods do not accept pathlib.Path() object as argument


roberto-arista:
hey @typemytype and @justvanrossum!

I realized that some methods from the DrawBot API accept pathlib.Path() objects, so I assumed that any method dealing with files on disk would do that as well. But a few of them do not.

Here I collected all the methods from the public API that deal with files in Drawbot (did I miss anything?):

```python
#!/usr/bin/env python3

# ------------------------------- #
# DrawBot and pathlib.Path object #
# ------------------------------- #

# -- Modules -- #
import drawBot as dB
from pathlib import Path

# -- Variables -- #
fontPath = Path('IBMPlexMono-Light.otf')
imagePath = Path('drawBot.png')

# -- Instructions -- #
if __name__ == '__main__':
    # Formatted String
    fs = dB.FormattedString()
    fs.font(fontNameOrPath=fontPath)
    fs.fallbackFont(fontNameOrPath=fontPath)
    print(fs.listOpenTypeFeatures(fontNameOrPath=fontPath))
    print(fs.listFontVariations(fontNameOrPath=fontPath))
    print(fs.listNamedInstances(fontNameOrPath=fontPath))
    dB.text(fs, (100, 100))

    # Text functions
    # dB.installFont(fontPath)
    # dB.uninstallFont(fontPath)
    dB.font(fontNameOrPath=fontPath)
    dB.fallbackFont(fontNameOrPath=fontPath)
    dB.listOpenTypeFeatures(fontNameOrPath=fontPath)
    dB.listFontVariations(fontNameOrPath=fontPath)
    dB.listNamedInstances(fontNameOrPath=fontPath)

    # image
    dB.image(path=imagePath, position=(0, 0))
    # dB.imageSize(path=imagePath)
    # dB.imagePixelColor(imagePath, (0, 0))
    # dB.imageResolution(path=imagePath)
    dB.numberOfPages(path=imagePath.with_suffix('.pdf'))

    # imageObject
    im = dB.ImageObject()
    # im.open(path=imagePath)

    # saving
    dB.saveImage(Path('pathlib.pdf'))
```

The ones commented out, raise an Exception if a Path() is provided.
They work fine if the Path() is converted to string

```python
#!/usr/bin/env python3

# ------------------------------- #
# DrawBot and pathlib.Path object #
# ------------------------------- #

# -- Modules -- #
import drawBot as dB
from pathlib import Path

# -- Variables -- #
[...]

PR #454, opened 2021-12-03
adds support for pathlib.Path objects for a few functions

roberto-arista:
in relation to issue #449


roberto-arista:
to be honest, I don't know why these older commits show up in this pull request. I forked the repo a while ago, today I realigned the existing fork with fetch upstream and then I made the edits to the code. Hopefully it is not a problem, if necessary I can start from scratch

drawBot/drawBotDrawingTools.py
```diff
@@ -1933,6 +1934,8 @@ def imageSize(self, path, pageNumber=None):
             # its an NSImage
             rep = path
         else:
+            if isinstance(path, pathlib.Path):
+                path = f'{path}'
```
justvanrossum, line 1938: 1. To convert a single object to a string one uses `str(x)`, not an f-string that only contains a single placeholder
2. However, the correct way to convert a Path object to a string path is `os.fspath(p)`
3. You need to do `isinstance(path, os.PathLike)` instead of checking agains `pathlib.Path`

drawBot/drawBotDrawingTools.py
```diff
@@ -2005,6 +2008,8 @@ def imagePixelColor(self, path, xy):
         x, y = xy
         if isinstance(path, str):
             path = optimizePath(path)
+        elif isinstance(path, pathlib.Path):
+            path = f'{path}'
```
justvanrossum, line 2012: Here, you should change the previous "if" into `if isinstance(path, (str, os.PathLine))`. `optimizePath()` calls `os.fspath()`, so no further action is needed here.

drawBot/drawBotDrawingTools.py
justvanrossum: Also, see comment below: change the `if isinstance(path, str)` line below to `if isinstance(path, (str, os.PathLike))` and remove the elif.

drawBot/context/tools/imageObject.py
```diff
@@ -56,17 +57,21 @@ def open(self, path):
         """
         if isinstance(path, AppKit.NSImage):
             im = path
-        elif isinstance(path, str):
-            path = optimizePath(path)
+        else:
```
justvanrossum, line 60: Same here, leave the original elif, extend isinstance with `os.PathLike`, get rid of the new else.

drawBot/drawBotDrawingTools.py
```diff
@@ -2275,6 +2286,10 @@ def uninstallFont(self, path):
             "uninstallFont(path) has been deprecated, use the font path directly in "
             "all places that accept a font name."
         )
+
+        if isinstance(path, pathlib.Path):
```
justvanrossum, line 2290: As above, replace with `path = os.fspath(path)` unconditionally.

justvanrossum (changes requested):

justvanrossum:
Maybe you fetched upstream after you created the PR branch?

roberto-arista:
> Maybe you fetched upstream after you created the PR branch?

I don't think so, but also I am not 100% sure. I am not very familiar with pull requests from forks

justvanrossum:
Can you please reread my feedback and ask if it's not clear?

I wrote `isinstance(path, (str, os.PathLike))` in some places, and `path = os.fspath(path)` unconditionally in the others.

I'm surprised the tests even pass right now, as you're not testing for `str` anymore.

roberto-arista:
Sorry, I thought os.PathLike worked as an umbrella for both. I misread your example.
I'll fix it on Monday

drawBot/drawBotDrawingTools.py
roberto-arista: so, if I understand correctly, in every spot where `optimizePath(path)` is called, I can use if `isinstance(path, (str, os.PathLike))`, right?

justvanrossum: I don't know about that, but in this place, yes.

roberto-arista: with "every spot", I meant every spot where I made the edits for this pull request. It's probably easier just to look at the modified code, I'll push it in a moment

justvanrossum (approved):

justvanrossum:
Perfect, thanks!

roberto-arista:
🎉

PR #457, opened 2021-12-06
add support for shareImage

typemytype:


drawBot/drawBotDrawingTools.py
```diff
@@ -462,6 +463,38 @@ def pdfImage(self):
...
+            # set A4 page size
+            size(200, 200)
+            # draw something
+            text("Foo, bar", (10, 10))
+            # share it over airdrop
+            shareImage('pdf', service="airdrop")
+        """
+        path = tempfile.mkstemp(suffix=f".{format}")[1]
+
+        self.saveImage(path, **kwargs)
```
justvanrossum, line 485: Sounds like a job for `saveImage("NSImage")`. The tempfile just keeps hanging around.

Also: shouldn't this be a proper menu item rather than a library function?

typemytype: the cool thing to have this as library function is that it can be run from code outside the app.

typemytype:
finally all test are green!

drawBot/drawBotDrawingTools.py
```diff
@@ -462,6 +463,39 @@ def pdfImage(self):
...
+        .. downloadcode:: shareImage.py
+
+            # set A4 page size
+            newPage(200, 200)
+            # draw something
+            text("Foo, bar", (10, 10))
+            # share it over airdrop
+            shareImage('pdf', service="airdrop")
+        """
+        path = tempfile.mkstemp(suffix=f".{format}")[1]
```
justvanrossum, line 483: Like I commented earlier, can we do this without a temp file? It's not being cleaned up, and I don't see how it could be. It's also not necessary as far as I can see: the sharing API works with all kinds of object, such as NSImage. Perhaps in-memory PDF data can be shared, too.

typemytype: A pdfDocument has no paste board support which is a requirement to send objects over. 

> The NSSharingService class is used to provide a consistent user experience when sharing items—NSURL objects, NSString objects, NSImage objects, video (through file URLs), or any object that implements the NSPasteboardWriting protocol—in macOS.

--> https://developer.apple.com/documentation/appkit/nssharingservice?language=objc

gif or mp4 has also no paste board support 

its indeed not cleaned up, as it needs to be available during the sharing process

I could add a NSSharingServiceDelegate where a callback is received when the sharing is done

https://developer.apple.com/documentation/appkit/nssharingservicedelegate?language=objc

justvanrossum: Then it indeed needs to be cleaned up with a delegate.

drawBot/drawBotDrawingTools.py
```diff
@@ -462,6 +463,39 @@ def pdfImage(self):
...
+
+        context = getContextForFileExt("pdf")
```
justvanrossum, line 485: Btw. why does this go via the pdf context and not via the regular `saveImage()`? Would this work for anything but pdf?

typemytype: cause it needs to be `format` this is wrong of me :)

justvanrossum: Question still stands: why not plain `saveImage()`?

drawBot/drawBotDrawingTools.py
typemytype: this is silly: the delegate removes the assets to fast and fe mail can not find it anymore...

see https://gist.github.com/typemytype/ea02152e754a5b0225df2df23976afd5

drawBot/drawBotDrawingTools.py
typemytype: must be `self.saveImage(..)` indeed

Issue #458, opened 2021-12-07
Setting tracking in a FormattedString causes DrawBot to crash


CalderH:
I have a program that adds data to a FormattedString, alternating between a header style (for people’s names) and a smaller style (for information about them). I add the small-text information by creating a separate FormattedString with that text, using that string’s .size() information to add line breaks in the right places, and then appending that FormattedString to the main FormattedString.

A specific person has some info that is slightly too wide for the text box, so (as a temporary solution) I added a case to test for whether it was them, and if so, decrease the tracking of their information before appending their FormattedString to the main one. However, adding the tracking-change line causes DrawBot to crash when the program runs. I haven’t found a way to reproduce the error in a simpler program (simply creating a FormattedString, appending some text to it, creating a second FormattedString with different tracking, appending some text to that, and then appending the second FormattedString to the first works fine).

If you’d like I can attach the file; I’ll just have to spend a moment editing out some inessential/possibly confidential info.

justvanrossum:
A small reproducer would be great!

CalderH:
Okay, I did manage to come up with a much smaller program that causes a similar error:
```
txt = FormattedString()
fake_database = [('', ''), ('One Two Three', 'Four Five Six')]
for data_row in fake_database:
    name, info = data_row
    txt += name + '\n'
    info_txt = FormattedString()
    if info == '':
        info_txt.tracking(-1)
        info_txt += 'Hello'
    else:
        info_txt += 'Hello'
    txt += info_txt
textBox(txt, (0, 0, 200, 200))
```
Removing the tracking line makes it no longer crash. Also, editing the string 'One Two Three' to 'One Two' makes it no longer crash, for some reason.

typemytype:
oh nice one, to simplify the example: adding tracking to an `\n` seems to fail somehow

```python
t = FormattedString()
t.fontSize(100)
t.tracking(-10)
t += "\n"
t.tracking(0)
t += "hello"

textBox(t, (30, 30, 300, 600))
```

CalderH:
Is adding tracking to `\n` necessarily the issue? I tried to separate the tracking adjustment from the `\n` by doing the following (in the original program), but it still crashed.

```
conc_txt.tracking(-1)
conc_txt += 'Social Innovation and Entrepreneurship |'
conc_txt.tracking(0)
conc_txt += '\n'
```

typemytype:
yep an other case:

```python
t = FormattedString()
t.tracking(-10)
t += "foo"
[...]

PR #459, opened 2021-12-07
only add tracking when the added text is not a \n or \r

typemytype:
fixing #458


typemytype:
this seems to be a bug deep in CoreText 

there is an endless loop when creating a `CoreText.CTFramesetterCreateFrame(setter, (0, 0), path, None)` with setter containing a attributed string with a tracked return

PR #471, opened 2022-02-28
Support post format-3 fonts in fs.appendGlyph()

justvanrossum:
Assuming fonttools-made glyph names.


justvanrossum:
Hm, fonttools actually does smarter glyph name synthesys for *encoded* glyphs (using AGL and uniXXXX names). Perhaps this change so far is too simplistic.

I don't think we have access to a (cached) TTFont instance to make a reverse mapping efficiently possible.

Alternatively, `fs.appendGlyph()` could accept glyph ids by means of an `isinstance(glyphName, int)` check.

typemytype:
Support indexes as `int` is perfect

typemytype:
+ adding test
+ fixing other tests :)

justvanrossum:
LGTM :)

drawBot/context/baseContext.py
```diff
@@ -1959,13 +1962,22 @@ def appendGlyph(self, *glyphNames):
...
+                glyph = glyphName
+            else:
+                glyph = font.glyphWithName_(glyphName)
             if glyph:
                 self.append(baseString)
                 glyphInfo = AppKit.NSGlyphInfo.glyphInfoWithGlyph_forFont_baseString_(glyph, font, baseString)
                 self._attributedString.addAttribute_value_range_(AppKit.NSGlyphInfoAttributeName, glyphInfo, (len(self) - 1, 1))
             else:
-                warnings.warn("font '%s' has no glyph with the name '%s'" % (font.fontName(), glyphName))
+                if isinstance(glyphName, int):
```
justvanrossum, line 1975: Not sure about this change: this basically checks whether the user passed 0 as a GID. Perhaps there should be a check after `glyphInfo = AppKit.NSGlyphInfo.glyphInfoWithGlyph_forFont_baseString_(glyph, font, baseString)`

typemytype: `.notdef` or `0` are both not possible. `.notdef` will result in glyph being `0` (it has to be the first glyph by index)

CoreText will fallback to an other font or the system font and will draw the `.notdef` of the systemfont which is a ? inside a diamond shape. And not the .notdef glyph of the current font...

typemytype: this produces also a different warning message based on the type of the `glypyhName`

justvanrossum: It is now possible to pass an out-of-range GID and cause an error. `glyphInfo` will be `None` in that case, and `self._attributedString.addAttribute_value_range_(...)` will throw something like this:

`ValueError: NSInvalidArgumentException - NSConcreteMutableAttributedString addAttribute:value:range:: nil value`

justvanrossum: And also, if you pass a GID of 0, yopu will get this warning:
```
message = "font '{fontName}' has no glyph with index '{glyphName}'"
```
Which makes no sense as the font _does_ contain a glyph with GID == 0.

PR #472, opened 2022-03-01
allow RF to work even if PIL is not available

### LINKED ISSUE #470: make the PILContext optionally only if PIL is installed
drawBot could be used in environments where PIL is not available.

https://github.com/typemytype/drawBotRoboFontExtension/issues/12

wrap the import in a try except ImportError and raise when a PIL context will be used

typemytype:
Only the PILContext will not work

fixes #470


typemytype:
test are fixed in the PR #471

justvanrossum (approved):

Issue #474, opened 2022-04-04
with context for newDrawing() / endDrawing()


roberto-arista:
👋

I find myself more and more running drawBot outside of the app, and I often encapsulate my code between `newDrawing()` and `endDrawing()`. Would you be open to add a `with` context as `with savedState()` to the API?

If yes, I could take care of opening a pull request using `savedState` as a reference : )

justvanrossum:
Yes.

What name do you propose?

```python
with newDrawing():
    ....
```

Or a new name?

roberto-arista:
I assume 

```python
with newDrawing():
    ...
```

could cause overshadowing issues, right? 

What about a simple
```python
with drawing():
    ...
```
?

Maybe too simple?

justvanrossum:
Or maybe even "Drawing" (cap-D)?

roberto-arista:
Yep!

typemytype:
+1 

why capitalise? it will not be a class name

justvanrossum:
Not sure this should be an argument, but in `drawbot-skia`, the place to inplement this is a class named "Drawing" :) https://github.com/justvanrossum/drawbot-skia/blob/master/src/drawbot_skia/drawing.py

justvanrossum:
But in terms of API is should probably just be a `contextlib.contextmanager` function, indeed not unlike `savedState()`. Perhaps "drawing" is fine, too.

PR #475, opened 2022-04-06
add `with drawing()` to public API

roberto-arista:
following this discussion: #474


drawBot/drawBotDrawingTools.py
```diff
@@ -180,6 +199,27 @@ def endDrawing(self):
...
+            # and it uses the `with drawing()` statement to ensure that page numbers
+            # restart from 1 in each PDF
+            for eachBooklet in range(1, 4):
+                with drawing():
+                    for eachPage in range(10):
+                        newPage(400, 400)
+                        text(f"{pageCount()}", (100, 100))
+                    saveImage(f"book_{eachBooklet}.pdf")
+        """
+        return DrawingContextManager(self)
```
justvanrossum, line 221: Suggest to use a contextlib.contextmanager decorator on this function instead:

- https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

Also, we'll need an "expected result" image for the demo code. I forgot what the easiest way to do that is.

roberto-arista:
Hey Just, I've read the docs for `contextlib.contextmanager` but I am a bit confused on how it applies to this case. The docs say 

> The function being decorated must return a [generator](https://docs.python.org/3/glossary.html#term-generator)-iterator when called. This iterator must yield exactly one value, which will be bound to the targets in the [with](https://docs.python.org/3/reference/compound_stmts.html#with) statement’s as clause, if any.

How can `drawing()` return a generator-iterator?

typemytype:
I will add the expected example test images

justvanrossum:
As per the example from the docs, it should look roughly like this:

```python
    contextlib.contextmanager
    def drawing(self):
        self.newDrawing()
        try:
            yield
        finally:
            self.endDrawing()
```

roberto-arista:
The code should be fine, now it seems I have to figure out the tests images

typemytype:
the same contextmanager could be used for `savedState`

roberto-arista:
👋

roberto-arista:
Hey all!
Is there anything I can do to push forward this pull request?
I think it would be a nice addition to the drawbot API!

roberto-arista:
🎉

justvanrossum:
> I could not open a separate pull request because this one has not been merged yet.

PR's don't need to be merged in order of creation, especially if their respective changes don't conflict.

And once one is merged, it is possible to rebase the other.

roberto-arista:
Wait, I thought this was merged, instead it was only approved. I'll reopen the pull request!

Issue #477, opened 2022-05-06
relative path for fonts triggered CoreText error in terminal


roberto-arista:
hey all!

let's take the following example:

```python3
from drawBot as dB
dB.newDrawing()
dB.newPage(1000, 1000)
dB.font("RobotoFlex[GRAD,XOPQ,XTRA,YOPQ,YTAS,YTDE,YTFI,YTLC,YTUC,opsz,slnt,wdth,wght].ttf")
dB.text("hello", (100, 100))
dB.saveImage("hello.pdf")
dB.endDrawing()
```

it works fine both in the app and from terminal.
If, instead, I have the font placed in a folder that I have to access relatively, like this:

```python3
from drawBot as dB
dB.newDrawing()
dB.newPage(1000, 1000)
dB.font("../font/RobotoFlex[GRAD,XOPQ,XTRA,YOPQ,YTAS,YTDE,YTFI,YTLC,YTUC,opsz,slnt,wdth,wght].ttf")
dB.text("hello", (100, 100))
dB.saveImage("hello.pdf")
dB.endDrawing()
```

in the terminal, it triggers the following traceback

```
2022-05-06 12:22:28.246 python[10359:104404] CoreText note: Client requested name "../font/RobotoFlex[GRAD,XOPQ,XTRA,YOPQ,YTAS,YTDE,YTFI,YTLC,YTUC,opsz,slnt,wdth,wght].ttf", it will get Times-Roman rather than the intended font. All system UI font access should be through proper APIs such as CTFontCreateUIFontForLanguage() or +[NSFont systemFontOfSize:].
2022-05-06 12:22:28.246 python[10359:104404] CoreText note: Set a breakpoint on CTFontLogSystemFontNameRequest to debug.
```

in the app works fine!

typemytype:
reproducible with python py3.10, not with py3.7... 

```python
import AppKit
AppKit.NSFont.fontWithName_size_("../relativePath/font.otf", 10)
```

instead of trying to retrieve the font by name first, check if the font name is an existing path.

see https://github.com/typemytype/drawbot/blob/5dd8e8781cf80771291c8de2c957b06ec263cfb1/drawBot/context/baseContext.py#L2645-L2664

typemytype:
the problem is that dot prefixed font names are system font and this seems not to be allowed when requesting a font with `AppKit.NSFont.fontWithName_size_`

typemytype:
[...]

PR #478, opened 2022-05-08
skip dot prefix font names

typemytype:
it must be a relative path

fixing #477


frankrolf:
Following up in the remark in #479: I hope those fonts will then still be visible in `installedFonts()` (Otherwise I wouldn’t know how to use SF Mono)

typemytype:
see #477 using dot prefixed font names doesnt work anyhow....

you can download SF Mono and install them from here https://developer.apple.com/fonts/

PR #484, opened 2022-09-07
Move to packaging version

typemytype:


setupApp.py
```diff
@@ -56,27 +55,6 @@ def _findModules(root, extensions, skip, parent=""):
                 else:
                     yield moduleName
 
-def getStdLibModules():
```
justvanrossum, line 59: Why this change?

typemytype: oh will revert... this is result of research done with the latest py2app and embedding std module. It seems not be necessary to explicitly list every module (will be a different PR)

PR #496, opened 2022-12-20
move to py3.9

typemytype:


requirements.txt
```diff
@@ -2,13 +2,12 @@ setuptools>=45.2.0
...
 corefoundationasyncio
-# pin py2app to version 0.19, see https://github.com/ronaldoussoren/py2app/issues/281
-# fix error: Failed to connect (colorGridView) outlet from (NSApplication) to (NSColorPickerGridView): missing setter or instance variable
-py2app==0.19
+py2app
 cocoa-vanilla==0.2.0
 defcon
 git+https://github.com/robotools/defconAppKit
 fontTools[ufo,lxml,woff,unicode,type1]
+uharfbuzz
```
justvanrossum, line 10: Why this new requirement?

typemytype: we where installing a custom wheel for harfbuzz all the time, without being mentioned in the requirements see https://github.com/typemytype/drawbot/blob/master/.github/workflows/build.yml#L54

justvanrossum: So it is rather a "bundled extra" than a real requirement I assume? It is not listed in setupApp.py.

If we want to bundle it (which I don't think is needed), we should also list it in setupApp.py.

In fact, if we don't pin it, we could only mention it in setupApp.py, just like the other bundled items. (Which IMO could all be removed, but people still don't seem to find the Install Python Packages menu...)

typemytype: its gone!

PR #499, opened 2022-12-27
Adapt std lib module finder pattern to Py 3.9

### LINKED ISSUE #498: missing module »cmath«
Hi all,
since the update to 3.129 (python3.10) it seems that the module »cmath« is missing.
When I'm using the pandas-tool (https://pandas.pydata.org) or just using `import cmath` an error Message pops up: »ModuleNotFoundError: No module named 'cmath'«

justvanrossum:
This is perhaps not the ideal general fix, but it does fix #498 for Python 3.9.


typemytype:
should we do a new release directly? 

if so version number must be touched and we should add something in the changelog

PR #509, opened 2023-02-14
Correct docstring

roberto-arista:
`textBoxCharacterBounds()` does not return `characters`

```
CharactersBounds(bounds=(0.0, 0.890625, 195.361328125, 11.77734375), baselineOffset=2.109375, formattedSubString=some text )
```


roberto-arista:
I don't think the failing tests are related to the docstring 🧐

typemytype:
it seems like Courier changes in 10.16... just fixed the test

typemytype:
thanks! for the doc fix :)

PR #519, opened 2023-03-10
Add supported formats to image functions docstrings

roberto-arista:


drawBot/drawBotDrawingTools.py
```diff
@@ -1902,7 +1902,7 @@ def FormattedString(self, *args, **kwargs):
     def image(self, path, position, alpha=1, pageNumber=None):
         """
         Add an image from a `path` with an `offset` and an `alpha` value.
-        This should accept most common file types like pdf, jpg, png, tiff and gif.
+        This should accept most common file types like pdf, jpg, png, tiff, gif and `NSImage`.
```
justvanrossum, line 1905: `NSImage` is not a file type, though, so this is a bit confusing.

roberto-arista: Something like:

> This should accept most common file types like pdf, jpg, png, tiff, and gif. `NSImage` objects are accepted too.

justvanrossum: Sounds good to me.

While at it, I would change "should accept" to "accepts".

justvanrossum (approved):

Issue #518, opened 2023-03-09
Combining `textBoxCharacterBounds()` with `FormattedString.fontFilePath()` does not seem to work


roberto-arista:
I am testing the [`FormattedString.fontFilePath()`](https://www.drawbot.com/content/text/formattedString.html#drawBot.context.baseContext.FormattedString.fontFilePath) method with a portion of the first code example of the `FormattedString` object.

If I run this code:

```python
fs = FormattedString()

size(1000, 200)
# create a formatted string
txt = FormattedString()

# adding some text with some formatting
txt.append("hello", font="Helvetica", fontSize=100, fill=(1, 0, 0))
# adding more text
txt.append("world", font="Times-Italic", fontSize=50, fill=(0, 1, 0))

# setting a font
txt.font("Helvetica-Bold")
txt.fontSize(75)
txt += "hello again"

# drawing the formatted string
text(txt, (10, 30))

bounds = textBoxCharacterBounds(txt, (0, 0, width(), height()))
for eachBox in bounds:
    print(eachBox.formattedSubString)
    print(eachBox.formattedSubString.fontFilePath())

```

I get the following result:

```
hello
/System/Library/Fonts/Helvetica.ttc
world
/System/Library/Fonts/Helvetica.ttc
hello again
/System/Library/Fonts/Helvetica.ttc
```

while I would expect something like this:

```
hello
/System/Library/Fonts/Helvetica.ttc
world
/System/Library/Fonts/Times-Italic.ttc
hello again
/System/Library/Fonts/Helvetica.ttc
```

PR #520, opened 2023-03-10
textProperties

typemytype:
textProperties returns a dict with all current properties of an formattedString. 

While building the nsAttributedString those properties are stored in the nsAttributedString with the key: `"drawBot.formattedString.properties"`. This allows to slice a formattedString and restore the attributes from the nsAttributedString.

The method is a public one as this can be used to check text properties after creating/building the formattedString.

fixing #518

### DIFF
--- a/docs/content/text/textProperties.rst
+++ b/docs/content/text/textProperties.rst
@@ -16,6 +16,7 @@ Text Properties
 .. autofunction:: drawBot.listNamedInstances
 .. autofunction:: drawBot.tabs
 .. autofunction:: drawBot.language
+.. autofunction:: drawBot.textProperties
 
 Font Properties
 ---------------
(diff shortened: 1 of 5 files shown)


typemytype:
@justvanrossum this would be super handy to have!!

justvanrossum:
Neat!

Could there be a need/use to set the properties from the dict as well? Or is this a read-only thing (from the API's perspective)?

typemytype:
you can use this properties dict to initiate a new `FormattedString("hi Just", **textProperties)` 

this super handy when you loop over all `textBoxCharacterBounds` to get properties and only change for the alpha value of the fill color... over an animation

justvanrossum:
And in between adding text?
```python
fs = Formattedtring(...)
fs.append("Hey ")
fs.???(**textProperties)
fs.append("Frederik")
```

typemytype:
works also with append

```python

fs = Formattedtring(...)
fs.append("Hey ", **textProperties)
```

justvanrossum (approved):

Issue #514, opened 2023-02-21
Improve `language()` support / implementation


roberto-arista:
Hey there!

I would like to improve documentation and feedback of the `language()` feature, both for text funcs and formatted strings. 

Reading some documentation from the apple developer website, it looks like CoreText accepts any ISO 639-2 or 639-1 language string abbreviation:
https://developer.apple.com/documentation/foundation/nslocale/1418015-isolanguagecodes?language=objc

It would be cool to warn the user if a string provided is not part of that list of languages.
Right now, code like this:
```
language("blablabla")
```
does not provide any feedback.

@justvanrossum @typemytype what do you think about this?

justvanrossum:
A warning sounds useful. Are these strings case sensitive or not? If not: make sure to take that into account when checking.

typemytype:
for now there is already a warning when a `language` is set which is not available: `*** DrawBot warning: Language 'foo' has no hyphenation available. ***` (only while drawing text)

```python
language("foo")
hyphenation(True)
text("bar", (10, 10))
```

a check is maybe a bit more complex then just looking if the language tag is in that list, fe: `nl-be` or `fr-CA`, this is the full spec: `language-extlang-script-region-variant-extension-privateuse` [see](https://www.w3.org/International/articles/language-tags/)

so `language()` could draw some dummy text somewhere that produce this warning..

typemytype:
had to look it up, this could be added while setting `language`: https://github.com/typemytype/drawbot/blob/6697d569d5a6b2c5b446083ea6cf79609c31e745/drawBot/drawBotDrawingTools.py#L1519

justvanrossum:
> had to look it up, this could be added while setting language:

But that checks whether there is hyphenation for that language, which may be distinct from "we have a valid language".

typemytype:
this list all possible options for `language(..)`
```python
import AppKit
print(AppKit.NSLocale.availableLocaleIdentifiers())
```

justvanrossum:
It may have to be something in this direction:
```python
import AppKit

def isLanguageCodeOk(langCode):
    parsedLoc = AppKit.NSLocale.componentsFromLocaleIdentifier_(langCode)
    lang = parsedLoc[AppKit.kCFLocaleLanguageCodeKey]
    country = parsedLoc.get(AppKit.kCFLocaleCountryCodeKey)
    return lang in AppKit.NSLocale.ISOLanguageCodes() and (
[...]

PR #521, opened 2023-03-10
check language

typemytype:
fixing #514

### DIFF
--- a/drawBot/drawBotDrawingTools.py
+++ b/drawBot/drawBotDrawingTools.py
@@ -17,7 +17,7 @@ from .context.tools import gifTools
 from .context.tools import openType
 from .context.tools import drawBotbuiltins
 
-from .misc import DrawBotError, warnings, VariableController, optimizePath, isPDF, isEPS, isGIF, transformationAtCenter, clearMemoizeCache
+from .misc import DrawBotError, warnings, VariableController, optimizePath, isPDF, isEPS, isGIF, transformationAtCenter, clearMemoizeCache, validateLanguageCode
 
 
 def _getmodulecontents(module, names=None):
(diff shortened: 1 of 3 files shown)


tests/testMisc.py
```diff
@@ -364,6 +364,28 @@ def test_remap(self):
         assert drawBot.remap(25, 10, 20, 30, 50) == 60
         assert drawBot.remap(25, 10, 20, 30, 50, clamp=True) == 50
 
+    def test_validateLanguageCode(self):
+        expectedLanguageValidation = [
+            ("ab", False),
+            ("abk", False),
+            ("abz", False),
+            ("AB", False),
```
justvanrossum, line 372: Some actual ISO 639-1 vs ISO 639-2 tests:
```suggestion
            ("AB", False),
            ("af", True),
            ("afr", True),
            ("af_NA", True),
            ("afr_NA", True),
            ("af_ZZ", False),
            ("afr_ZZ", False),
```

drawBot/drawBotDrawingTools.py
```diff
@@ -1480,8 +1480,9 @@ def tabs(self, *tabs):
     def language(self, language):
         """
         Set the preferred language as language tag or None to use the default language.
-
-        Support is depending on local OS.
+        A language tag might be a [iso639-2 or iso639-1](https://www.loc.gov/standards/iso639-2/php/English_list.php)
+        code or a locale identifier supported by local OS.
+        A warning will be raised if the language tag is not supported.
```
justvanrossum, line 1485:
```suggestion
        A warning will be issued if the language tag is not supported.
```

roberto-arista:
✨

PR #536, opened 2023-07-07
Update `linkRect()` example with snippet by @typemytype

roberto-arista:


roberto-arista:
Test on linkRect is skipped apparently

Screenshot 2023-07-10 at 10 44 47

roberto-arista:
https://github.com/typemytype/drawbot/blob/4313294e57bd02e7d384bfd313d49e55977edba9/tests/testExamples.py#L160-L164

roberto-arista:
👏

PR #540, opened 2023-08-25
get the correct image size when an nsimage object is provided

typemytype:


drawBot/drawBotDrawingTools.py
```diff
@@ -1936,7 +1936,13 @@ def imageSize(self, path, pageNumber=None):
 
         if isinstance(path, AppKit.NSImage):
             # its an NSImage
```
justvanrossum, line 1938:
```suggestion
            # it is an NSImage
```

justvanrossum (approved):

Issue #541, opened 2023-09-20
set CGContextSetAllowsFontSubpixelQuantization to False


typemytype:
during animation with `text` and `textBox` minimal changes in the formattedString (pointSize, variable settings, ...) influences the pixel output heavily 

https://github.com/typemytype/drawbot/assets/1190358/ba71ba1e-9474-473e-9f91-8116c5a7fa8f

setting `Quartz.CGContextSetAllowsFontSubpixelQuantization(context, False)` disables subpixel rendering, [see](https://developer.apple.com/documentation/coregraphics/1456263-cgcontextsetallowsfontsubpixelqu)

https://github.com/typemytype/drawbot/assets/1190358/aa85da13-0e45-4e14-8123-d02c1658ea13

proposal:

add `saveImage` argument: `subpixelRendering=False` the default would be `True` so nothing changes

justvanrossum:
Great find! No longer we have to render text as BezierPaths to avoid this effect!

PR #542, opened 2023-09-20
Subpixel rendering

typemytype:
fixing #541


justvanrossum (commented):
Let's call the flag fontSubpixelQuantization=False/True to stay closer to the original and be more precise in what it actually does. It's about the positioning, not the rendering, which will use subpixels in both settings.

typemytype:
`fontSubpixelQuantization` it is!

PR #545, opened 2023-11-10
Update `linkRect()` example with snippet by @typemytype

roberto-arista:
I made these edits a while ago and I forgot to merge them into the repo. What do you think @typemytype?


typemytype:
looks idd beter... and the linkRect example is excluded in the tests ([see](https://github.com/typemytype/drawbot/blob/master/tests/testExamples.py#L163C5-L163C43))

typemytype:
will be in the website documentation when we tag a new release

Issue #531, opened 2023-06-21
CMYK Colors and Color profile do not export correctly with saveImage()


BennZorn:
Hey there! 

I wanted to export a .pdf or . jpeg or .tiff with cmyk values but the CMYK values in the exported file do not match the ones I set in the DrawBot code. Also, the Color Profile (in the datatypes where saveImage() supports Color Profiles) is not embedded, but Generic RGB. 

[DrawBotIssue-CMYK-ExportAndProfile.zip](https://github.com/typemytype/drawbot/files/11815658/DrawBotIssue-CMYK-ExportAndProfile.zip)

Issue #284, opened 2019-07-18
Export TIFFs with CMYK color


andyclymer:
Frederik, Just and I were discussing this so I'll open a feature — it could be nice to be able to specify that a TIFF image be exported in the CMYK colorspace.

Frederik noticed that this line is blocking it:
https://github.com/typemytype/drawbot/blob/2486f44a553f81870c07ad033fbfcfbcd3882028/drawBot/context/imageContext.py#L119

justvanrossum:
`samplesPerPixel` need to be 5 is `colorSpaceName` is `AppKit.NSDeviceCMYKColorSpace`.

Perhaps more arguments for `initWithBitmapDataPlanes_pixelsWide_pixelsHigh_bitsPerSample_samplesPerPixel_hasAlpha_isPlanar_colorSpaceName_bytesPerRow_bitsPerPixel_` should become arguments to `_makeBitmapImageRep`. At least `samplesPerPixel`, perhaps `hasAlpha`.

PR #554, opened 2024-01-21
Add support for saving CMYK images

fhoehl:
I needed to export TIFF images with CMYK. Even though I was setting `imageColorSyncProfileData` with macOS's Generic CMYK profile, the final output file did not have the colour profile set, and the encoded colours were in the RGB space.

I would happily invest more time and add tests if this PR looks valuable for the project!

This PR is related to #284 and #531.

### DIFF
--- a/drawBot/context/baseContext.py
+++ b/drawBot/context/baseContext.py
@@ -392,7 +392,6 @@ class BezierPath(BasePen, SVGContextPropertyMixin, ContextPropertyMixin):
         path, (x, y) = context._getPathForFrameSetter(box)
         attributedString = context.attributedString(txt, align)
 
-        print("hip", hyphenation)
         if hyphenation:
             attributedString = context.hyphenateAttributedString(attributedString, path)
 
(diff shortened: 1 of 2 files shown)


justvanrossum:
Hi, thank you so much for starting this! We are certainly interested in integrating this, and we would like you to encourage to write some tests. It is possible this will turn out harder than it should be. Let us know if you get blocked, then we can try to sort this out.

The "hyphenation" change in baseContext.py seems out of place. We would prefer a separate PR for that.

justvanrossum:
You never responded to the hyphenation feedback: what does that have to do with this PR?

There are test scripts in `tests/drawBotScripts/`, these tend to produce images, which are verified against the contents of `tests/data/`. It would be great if you could add that.

fhoehl:
@chrisjansky, I'm glad you find this PR useful. I must admit that I forgot about this PR, and a year has passed 😄. Let me see if I can add the test this week!

@justvanrossum, thanks for the pointers. I will remove the hyphenation, which is totally unrelated.

PR #568, opened 2024-03-04
add support for writingDirection(..)

### LINKED ISSUE #567: Arabic type settings mixed with Latin characters looks to be wrong
In Drawbot if the Arabic text starts with a Latin word, the Latin word appears at the wrong place, here on the left, but it should be on the right.
image
In text edit at first it behave the same

image
but if we change the writing direction to right to left, then the text is correct

image
As far as I know, in Drawbot looks to be impossible to change the writing direction.

typemytype:
This fixes #567.

No test yet.


typemytype:
fixing #567

justvanrossum:
@typemytype I added a test and a doc example. I think this is then good to go!

typemytype:
I guess testing only the doc string code is more then enough.

justvanrossum:
You're right, I'll remove the "normal" test. (I only thought of the doc test after I made the normal one...)

PR #570, opened 2024-05-17
[WIP] Type annotations

roberto-arista:
The aim of this pull request is to add type annotations to the `drawBot` public interface. Here are the main things @typemytype and I worked on:

- Adding annotations and fix the bugs the annotations exposed
- Adding annotations to the tools/imageObject.py module did not seem clever, since it's automatically generated from the Core Image Filters. @typemytype found the original Python script on his drive and we updated it to support annotations and add general improvements.
- On the same line, the `drawBot.__init__.py` file is partially generated to expose the instance methods outside, allowing `from drawBot import whatever` and benefiting from the annotations in the code editor

Additionally, we made some other minor fixes:
- Removed wildcard imports from `drawBot` test scripts (type annotations benefit)
- Updated the image URL (from Cloudflare to GitHub) and related test images
- Updated the README

Some tests are failing on my OS (Sonoma 14.4.1). Looking at the `difference.pdf`, it mostly seems to be related to hinting, hyphenation, text baseline shift, and blending of colors. Depending on the OS GitHub will use to run the tests, the tests might keep failing or not.

We still need to generate automatically basic tests for the `ImageObject` class. We plan to work on it next week.


drawBot/aliases.py
```diff
@@ -0,0 +1,12 @@
...
+
+Point = Size = tuple[float, float]
+Points = Sequence[Point]
+SomePath = os.PathLike | str
+CMYKColor = float | tuple[float, ...]
+# used for output values which are always 4 elements
+RGBAColorTuple = tuple[float, float, float, float]
+RGBColor = float | tuple[float, ...]
+BoundingBox = tuple[float, float, float, float]
+Transform = tuple[float, float, float, float, float, float]
```
justvanrossum, line 12: Should we consider calling this `TransformTuple`, do make it easier to distinguish from fontTools' Transform (which is a named tuple, and can be used here, too)?

justvanrossum (commented):
Great stuff!

I have some minor comments/suggestions.

I am missing a mypy check step in the GH Actions workflow files. How do you currently check the correctness?

drawBot/__init__.py
```diff
@@ -1,3 +1,124 @@
...
+underline = _drawBotDrawingTool.underline
+uninstallFont = _drawBotDrawingTool.uninstallFont
+url = _drawBotDrawingTool.url
+width = _drawBotDrawingTool.width
+writingDirection = _drawBotDrawingTool.writingDirection
+
+from drawBot.context.tools import drawBotbuiltins
+lerp = drawBotbuiltins.lerp
+norm = drawBotbuiltins.norm
+remap = drawBotbuiltins.remap
```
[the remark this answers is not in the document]
roberto-arista: I use Black (or Ruff) almost everywhere in my projects. I'd be happy to apply it here too. @typemytype would you like to experiment with this in a separate PR?

roberto-arista:
> Great stuff!

Thanks : )

> I am missing a mypy check step in the GH Actions workflow files. How do you currently check the correctness?

I use the Pylance extension in VSCode. I have no experience in setting it into GH Actions. There's a ton of errors we are just ignoring, I think it would be problematic to run it outside a code editor...

drawBot/drawBotDrawingTools.py
```diff
@@ -2311,7 +2402,7 @@ def textSize(self, txt, align=None, width=None, height=None):
             raise DrawBotError("Calculating textSize can only have one constrain, either width or height must be None")
         return self._dummyContext.textSize(txt, align, width, height)
 
-    def installedFonts(self, supportsCharacters=None):
+    def installedFonts(self, supportsCharacters: str | None = None):
```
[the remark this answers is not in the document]
roberto-arista: good catch!

drawBot/context/baseContext.py
```diff
@@ -167,7 +175,7 @@ def __repr__(self):
 
     # pen support
 
-    def moveTo(self, point):
+    def moveTo(self, point: Point):
```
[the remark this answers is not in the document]
roberto-arista: From what I can see on Pylance, that's redundant.

justvanrossum:
> I think it would be problematic to run it outside a code editor...

That would make the project depend on VS Code, which is highly undesirable. We need a way to guarantee the type annotations stay correct without VS Code. That is really is a hard requirement.

I have never heard of pylance, but we should either use it to do checking in GH Actions (and possibly ingore certain errors, if we know for sure they are not relevant to us), or we should use mypy. I have experience with the latter, is easy-ish to set up, but can be quite picky in terms of type checking.

I just read that pylance is powered by [pyright, which is available on PyPI](https://pypi.org/project/pyright/).

If pyright is easier than mypy, I'm totally fine with that, but we're going to have to use either of those as part of automated testing.

drawBot/__init__.py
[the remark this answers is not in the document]
typemytype: new PR, this is already a super big PR, and this seems more or less just aesthetic white space improvements

roberto-arista:
> > I think it would be problematic to run it outside a code editor...
> 
> That would make the project depend on VS Code, which is highly undesirable. We need a way to guarantee the type annotations stay correct without VS Code. That really is a hard requirement.
> 
> I have never heard of pylance, but we should either use it to do checking in GH Actions (and possibly ignore certain errors, if we know for sure they are not relevant to us), or we should use mypy. I have experience with the latter, is easy-ish to set up, but can be quite picky in terms of type checking.
> 
> I just read that pylance is powered by [pyright, which is available on PyPI](https://pypi.org/project/pyright/).
> 
> If pyright is easier than mypy, I'm totally fine with that, but we're going to have to use either of those as part of automated testing.

I agree on the fact that the codebase should be IDE agnostic. The problem is that we're just annotating a very small part of it. For example, on `drawBotDrawingTools.py` I now see 72 type errors that I am ignoring. So, if we want to add some automatic checking on GH Actions, it should be narrowed to a portion of the codebase.

drawBot/__init__.py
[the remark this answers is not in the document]
roberto-arista: I'll open an issue and resolve this one.

justvanrossum:
> it should be narrowed to a portion of the codebase

Mypy is pretty good with ingnoring things that aren't annotated.

justvanrossum:
[mypy_errors.txt](https://github.com/typemytype/drawbot/files/15376454/mypy_errors.txt)

justvanrossum:
FWIW, running `pyright drawBot --skipunannotated` gives:
```
630 errors, 8 warnings, 741 informations
```

drawBot/aliases.py
roberto-arista: fixed with c8c13714728abc1f7ace2662a104c50181d9d08e

drawBot/context/tools/imageObject.py
```diff
@@ -119,14 +130,14 @@ def unlockFocus(self):
         pageCount = data.pageCount()
         page = data.pageAtIndex_(pageCount - 1)
         # create an image
-        im = AppKit.NSImage.alloc().initWithData_(page.dataRepresentation())
+        im = AppKit.NSImage.alloc().initWithData_(page.dataRepresentation()) # type: ignore
```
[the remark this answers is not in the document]
roberto-arista: I don't remember why I added the comment, but all AppKit is dynamic, so static analysis does not work.

drawBot/drawBotDrawingTools.py
```diff
@@ -1239,7 +1321,7 @@ def lineCap(self, value):
         self._requiresNewFirstPage = True
         self._addInstruction("lineCap", value)
 
-    def lineDash(self, *value, offset=0):
+    def lineDash(self, *value: list[float], offset=0):
```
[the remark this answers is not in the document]
roberto-arista: You are right, but looking at the code underneath I am bit confused by these lines:
https://github.com/typemytype/drawbot/blob/f115a0ee051328efaf8bbe17e40f9f08dc3f81d1/drawBot/drawBotDrawingTools.py#L1282-L1283

Was it an older interface? If yes, should we add a warning? cc @typemytype

drawBot/drawBotDrawingTools.py
```diff
@@ -1684,19 +1761,19 @@ def fontNamedInstance(self, name, fontNameOrPath=None):
...
-    def textProperties(self):
+    def textProperties(self) -> dict[str, Any]:
         return self._dummyContext._state.text.textProperties()
 
     textProperties.__doc__ = FormattedString.textProperties.__doc__
 
     # drawing text
 
-    def text(self, txt, position, align=None):
+    def text(self, txt: _formattedStringClass | str, position: Point, align: str | None = None):
```
justvanrossum, line 1776: This causes an error when running the code (and mypy also complains about it):
```
NameError: name '_formattedStringClass' is not defined. Did you mean: 'FormattedString'?
```

drawBot/aliases.py
```diff
@@ -0,0 +1,12 @@
+import os
+from typing import Sequence
+
+Point = Size = tuple[float, float]
```
justvanrossum, line 4:
```suggestion
Point = tuple[float, float]
Size = tuple[float, float]
```
This makes mypy a lot happier.

drawBot/drawBotDrawingTools.py
roberto-arista: Yes, there's a conflict with the function name. Pyright and mypy give different errors, I'll look for a solution.

drawBot/drawBotDrawingTools.py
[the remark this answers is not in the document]
justvanrossum: Oh hmm, it seems to accept both `lineDash([a, b, c])` as well as `lineDash(a, b, c)`. If we want to keep this, we should annotate as:

```python
    def lineDash(self, *value: float | list[float], offset: float = 0):
```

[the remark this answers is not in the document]
justvanrossum: We don't verify that len(value) == 1 in the case of `lineDash([a, b, c])`. That case would warrant a TypeError.

We may want to deprecate the list spelling, but let's consider that a separate issue.

drawBot/drawBotDrawingTools.py
roberto-arista: Both type checkers are having a hard time with the function name and the class name being the same. I've opted for a different solution in 7a3199647ccb6f6ab05c7d817277e641920e021a. For the sake of alignment, we could do the same to `BezierPath` and `ImageObject`.

roberto-arista:
Files to include in the typechecking:
- `drawBot/__init__.py`
- `drawBot/aliases.py`
- `drawBot/drawBotPackage.py`
- `drawBot/context/baseContext.py`
- `drawBot/context/tools/imageObject.py`
- `drawBot/context/tools/drawBotbuiltins.py`
- `drawBot/drawBotDrawingTools.py`

I've modified my setup to work with `mypy`. I'll start to go through the errors, highlight what needs to be fixed and ignore what's not feasible.

drawBot/drawBotDrawingTools.py
justvanrossum: I wonder if we should get rid of those "method aliases to class objects", and populate the namespace with the proper classes instead. I never liked the fact that "FormattedString" is a method on the context instead of the intended class.

typemytype: I agree to move BezierPath and ImageObject to proper class objects, but out of scope of this PR

roberto-arista: MyPy does not complain about this, so for now I'd wait for a new PR.

roberto-arista:
@typemytype and I worked today to fix the typing issues in the public interface. MyPy has been set as a step in the test workflow, so we almost ready! We plan to add some very basic tests for the `ImageObject` methods next week.

roberto-arista:
It took a while, but finally all tests are passing : )

typemytype:
🙌 super!! thanks

roberto-arista:
thank you for all the support and guidance

typemytype:
this looks good to me!

sphinx doc building is also adjusted and improved

roberto-arista:
How is it going with the review process? @typemytype @justvanrossum

justvanrossum:
> How is it going with the review process?

Apologies! I will continue soon.

drawBot/context/svgContext.py
```diff
@@ -35,7 +35,7 @@ def _intGeneratorFunc():
 # simple file object
 
 
-class SVGFile(object):
+class SVGFile():
```
justvanrossum, line 38: I think I commented this before: please remove the empty `()` in class definitions: this is uncommon style, and black would delete it.

drawBot/__init__.py
```diff
@@ -1,3 +1,125 @@
 from .drawBotDrawingTools import _drawBotDrawingTool
 
 _drawBotDrawingTool._addToNamespace(globals())
+
+
+# This section is automatically generated with scripting/generateDrawbotInit.py
```
justvanrossum, line 6: Would be good to have a test that verifies that the generated module still matches the current module, so we can't forget when it needs to be done.

drawBot/drawBotDrawingTools.py
```diff
@@ -995,7 +1056,13 @@ def cmykShadow(self, offset, blur=None, color=None):
         self._requiresNewFirstPage = True
         self._addInstruction("cmykShadow", offset, blur, color)
 
-    def linearGradient(self, startPoint=None, endPoint=None, colors=None, locations=None):
+    def linearGradient(
+        self,
+        startPoint: Point | None = None,
+        endPoint: Point | None = None,
+        colors: list[RGBColor] | None = None,
```
justvanrossum, line 1063: `RGBColor` is defined as `float | tuple[float, ...]`, but a list of floats does not work here.

drawBot/context/__init__.py
```diff
@@ -55,14 +55,16 @@ def getFileExtensions():
...
-def getContextOptionsDocs(formatter="* `%s`: %s"):
+def getContextOptionsDocs(formatter="* `%s`: %s") -> list[str]:
     docs = []
     for context in allContexts:
         if context.saveImageOptions:
             ext = ", ".join(context.fileExtensions)
             docs.append("*%s options:*" % ext)
             docs.append("")
-            for key, doc in context.saveImageOptions:
+            for option in context.saveImageOptions:
```
justvanrossum, line 65: I wonder if this change was triggered by the type annotation for saveImageOptions, which I think is wrong. I'll comment there.

drawBot/context/baseContext.py
```diff
@@ -2143,8 +2322,8 @@ class BaseContext(object):
     _bezierPathClass = BezierPath
     _gradientClass = Gradient
 
-    fileExtensions = []
-    saveImageOptions = []
+    fileExtensions: list[str] = []
+    saveImageOptions: list[str] = []
```
justvanrossum, line 2326: Code using this generally writes `for a, b in xxx.saveImageOptions`, and I guess the type annotation should be `list[tuple[str, str]]`

justvanrossum (commented):
I started to do a new review, and left some comments. But then I came across my earlier feedback which seems to have not been addressed at all, so I'll stop here.

drawBot/context/tools/imageObject.py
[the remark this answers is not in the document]
justvanrossum: That doesn't answer my question, though.

roberto-arista:
I'll go through them on Monday!

drawBot/drawBotDrawingTools.py
roberto-arista: Mmm, I understand what you mean. Colors should be tuples of 3 or 4 floats, correct?
Then we need to change the docstring too as 
> * `colors` as a list of colors, described similary as `fill`
it's not entirely correct

drawBot/context/baseContext.py
roberto-arista: fixed with b3bc4777f1b748bfbcccf8c0ffad462379f58fb8

roberto-arista: mmm, I reverted the change. The annotations do not align correctly.

drawBot/drawBotDrawingTools.py
roberto-arista: fixed with 1bdd28d3da63a9a7b50d4de304cfab5efbe08e5d

drawBot/__init__.py
roberto-arista: good idea!

roberto-arista: fixed with f9276587a7d89db9b47387acdcb19deaea2b08f6

drawBot/context/__init__.py
roberto-arista: fixed with 49c19855df09dc1c315f74bc78e5c69496a60f66

roberto-arista:
Hey @justvanrossum! @typemytype and I fixed the issues you highlighted. You can proceed with the review 🤓

typemytype:
@justvanrossum I would like to finish this PR :)

justvanrossum:
> I would like to finish this PR :)

I can imagine! :)

There's still a bunch of unresolved feedback, though. (And some feedback that has been resolved, but not marked as resolved.)

roberto-arista:
I think I resolved all the issues. There are commit links too. Am I missing something?

justvanrossum:
Just go to “show hidden” and within that one more time.

drawBot/context/baseContext.py
```diff
@@ -1485,105 +1537,139 @@ def fallbackFont(self, fontNameOrPath, fontNumber=0):
...
-    def fontSize(self, fontSize):
+    def fontSize(self, fontSize: float):
         """
         Set the font size in points.
         The default `fontSize` is 10pt.
         """
         self._fontSize = fontSize
 
-    def fill(self, *fill):
+    def fill(self,
```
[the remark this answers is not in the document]
roberto-arista: I'd leave the formatting with Black to a future PR.

typemytype:
To be honest, looking to outdated hidden message is a bit confusing... but hurray found them deep inside the comment stream!

drawBot/context/tools/imageObject.py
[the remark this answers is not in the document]
roberto-arista: Apparently these `# type: ignore` after `AppKit` module calls are not necessary since we are ignoring it directly at the import statement. Fixed with c41c63b5f825d4f38bcd4939d1ea815e6ca1555d

drawBot/context/baseContext.py
```diff
@@ -638,7 +687,7 @@ def removeOverlap(self):
         self.setNSBezierPath(result.getNSBezierPath())
         return self
 
-    def difference(self, other):
+    def difference(self, other: Self):
```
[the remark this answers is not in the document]
roberto-arista: Improved with bb5d6a487871a70e0722754770bdc82ed4699d21

drawBot/drawBotDrawingTools.py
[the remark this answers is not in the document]
roberto-arista: This section changed quite a bit since your comment. We update the code and the method signature. It looks like this now:
```python
def lineDash(self, value: float | None, *values: float, offset: float = 0):
    combinedDashes: list[float | None] = [value]
    if isinstance(value, float | int):
        combinedDashes.extend(values)
    self._requiresNewFirstPage = True
    self._addInstruction("lineDash", combinedDashes, offset)
```

roberto-arista:
@justvanrossum I should have answered to all the issues. Please, if there's something missing tag me directly, as the PR interface is getting a bit confusing, thanks!

roberto-arista:
Hey! Can I do anything to move the ball forward?

roberto-arista:
Hey hey! Let me do if I can do anything to move the ball forward : )

roberto-arista:
Hey hey, I recently pushed a couple of little improvements. I am ssing this fork in production in a couple of projects and I'd really love to see it merged upstream. 👋

justvanrossum:
I’ll try to get back to this soon!

RafalBuchner:
I know there are probably a lot of different priorities, but is there any perspective for this merge to happen?

roberto-arista:
Hey Rafal! Meanwhile you can use my fork, I've been using it in production for a while now.

justvanrossum:
Sorry for my sluggishness. @typemytype: maybe we should just merge, since all tests seem to pass? We can refine later if we run into issues.

typemytype:
oke, I will schedule some time this week

roberto-arista:
if you want some company @typemytype let me know : )

PR #579, opened 2025-02-12
improving test for imageObjects and all their filters

typemytype:
applying all filters during the test

ignoring lost of filters with colorspaces, camera, model input values

@roberto-arista


typemytype:
going to merge an touch a new version as non of the imageObject filter work...

PR #580, opened 2025-04-08
Lint codebase with `ruff` (and substitute `black`)

roberto-arista:
Hey!

Here's a PR with my implementation proposal for ruff in the drawBot codebase. In a nutshell:
- it adds a `lint.yml` file with instructions to format the code and auto-commit it back to the codebase
- therefore `test.yml` needs to start once the auto-commit is completed (workflow_run section)
- it adds a `ruff.toml` configuration file
- it adds the `ruff-api` package that allows calling `ruff` from the python codebase to substitute `black` (application menu > python > format code)

I've had mixed results with triggering one action from another when writing back some changes in the repo, we might explore different options eventually. Also, the ruff-api would like to have a filename to format the code (I guess for imports?), have a look at the FIXME here: drawBot/ui/drawBotController.py

That's it, let me know what you think!

👋


justvanrossum:
Thanks!

Not a big fan of autocommit, and I would prefer a pre-commit based approach that does apply fixes. What I do in other projects:

- Use pre-commit (locally), that runs black and flake8, which will fix issues, but aborts the commit if it had to do anything
- Use pre-commit also as a GHA step, that will fail if it has any complaints

Ideally I would like to see two separate PRs:
1. Reformat + lint the entire code base
2. Set up GHA / pre-commit to make sure the code base stays nicely formatted and lint-clean

roberto-arista:
on my way : )

app/ensure_universal_wheels.py
```diff
@@ -71,9 +71,7 @@ def merge_wheels(url1, url2, dest_dir):
         platform_base_parts = platform.split("_")
         platform_base = "_".join(platform_base_parts[:3])
 
-        universal_wheel_path = (
-            dest_dir / f"{wheel_base.lower()}-{platform_base}_universal2.whl"
-        )
+        universal_wheel_path = dest_dir / f"{wheel_base.lower()}-{platform_base}_universal2.whl"
```
justvanrossum, line 74: This file was previously formatted with black. Did you increase the max line length? Oh I see there's a config for 120 chars. Did you pick that 120 chars setting? If so, what was your reasoning? Not per se against, but there's some advantage to be compatible with default black settings.

drawBot/drawBotDrawingTools.py
```diff
@@ -1,39 +1,49 @@
...
 
-from .context import getContextForFileExt, getFileExtensions, getContextOptionsDocs
-from .context.baseContext import (
-    makeTextBoxes,
-    getNSFontFromNameOrPath,
-    getFontName,
-)
-from .context.baseContext import FormattedString, BezierPath
-from .context.dummyContext import DummyContext
+import AppKit  # type: ignore # type: ignore
```
justvanrossum, line 8: Not sure why this PR introduces these "type: ignore" comments in the first place, but this one is a bit much.

ruff.toml
```diff
@@ -0,0 +1,3 @@
+line-length = 120
+target-version = "py311"
+indent-width = 4
```
justvanrossum, line 3: It's somewhat ironic that the ruff config file is poorly formatted: it misses an ending newline :)

justvanrossum (commented):
I left a few minor comments. LGTM, apart from the test failure.

drawBot/drawBotDrawingTools.py
roberto-arista: maybe the import sorting of ruff caused this...

ruff.toml
roberto-arista: fixed and removed the indent-width since it's the default value

app/ensure_universal_wheels.py
roberto-arista: Yep, I've been working with 120 for quite a long time, I find 88 to be particularly short compared to modern screens. Also, func/methods definition with hints become quite long, and 88 feels even shorter in that kind of situation.

roberto-arista: I am curious about @typemytype opinion on the matter : )

roberto-arista:
I'll wait for this one to be merged before working on the github actions logic.

app/ensure_universal_wheels.py
justvanrossum: I think 120 fine for the DrawBot codebase.

justvanrossum (approved):
LGTM

justvanrossum:
@typemytype, shall we merge soon? I fear it will cause merge conflicts if we don't. Case in point: #582.

typemytype:
Sure, this is awesome! Is there a PR merge order?

justvanrossum:
> Is there a PR merge order?

No... In this case, If we merge PR A first, PR B will (likely) have a conflict. Maybe merge this one first.

roberto-arista:
im ooo, but I agree with Just, let's merge the linted code first, I'll work on the hook and workflow next week : )

justvanrossum:
(Oooh we got lucky: there was no merge conflict!)

typemytype:
thanks @roberto-arista !!

roberto-arista:
ty!

PR #582, opened 2025-04-25
Make sure we specify every axis in the font to ensure we get the expected default value behavior

### LINKED ISSUE #581: `db.text()` using a variable font does not show the neutral by default
If I open a variable font in FontGoggles (for example), it shows by default the neutral instance:

Image

Which, in the case of ABCGintoPlusVariable.ttf is `{'wdth': 100.0, 'wght': 400.0, 'slnt': 0.0, 'NORD': 0.0}`.

If I typeset the same file with drawBot, it shows instead what I suspect to be the first named instance (`'ABCGintoPlusVariable-CondensedHairline': {'wdth': 60.0, 'wght': 100.0, 'slnt': 0.0, 'NORD': 100.0}`):

Image

Then I've also noticed that if I print the result of `fontVariations()` after choosing the variable font with `font()` it returns correctly the neutral state (even though it does not typeset it on the canvas)

Here's the code I've been using:

```python
import drawBot as db

if __name__ == "__main__":
    with db.drawing():
        db.newPage(600, 400)

        db.fill(1, 1, 1)
        db.rect(0, 0, db.width(), db.height())

        db.font("ABCGintoPlusVariable.ttf", 75)
        db.fill(0, 0, 0)

        print(db.listFontVariations())
        print(db.listNamedInstances())
        print(db.fontVariations())

        db.text("ABC abc 0123 :;?", (100, 100))
        db.saveImage("neutral.png")

```

Attached here a subset font:

[ABCGintoPlusVariable.subset.ttf.zip](https://github.com/user-attachments/files/19887176/ABCGintoPlusVariable.subset.ttf.zip)

justvanrossum:
Somehow CoreText has different ideas what default to use if we don't specify an axis value. This PR reverses the logic of `getFontVariationAttributes()` and always sets all axis values.

This fixes #581.

### DIFF
--- a/drawBot/context/tools/variation.py
+++ b/drawBot/context/tools/variation.py
@@ -92,14 +92,17 @@ def getNamedInstancesForFont(font):
 def getFontVariationAttributes(font, fontVariations):
     coreTextFontVariations = dict()
     axes = getVariationAxesForFont(font)
+
     for axisTag, axis in axes.items():
         value = min(
             max(fontVariations.get(axisTag, axis["defaultValue"]), axis["minValue"]),
             axis["maxValue"],
         )
         coreTextFontVariations[convertVariationTagToInt(axisTag)] = value
+
     for axisTag in sorted(set(fontVariations) - set(axes)):
         warnings.warn(
             "variation axis '%s' not available for '%s'" % (axisTag, font.fontName())
         )
+
     return coreTextFontVariations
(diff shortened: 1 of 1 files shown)


typemytype:
Looks good!!

roberto-arista:
👏

roberto-arista:
The only doubt I have rn is that maybe we should test this output. I can propose something next week.

justvanrossum:
If your theory is correct that it picks the first named instance, it should be possible to whip up one-axis, one-glyph test font to reproduce the original issue.

PR #588, opened 2025-05-20
Use kCTTypesetterOptionAllowUnboundedLayout option for setting 'longer' strings, ensuring OT features are applied in all cases

### LINKED ISSUE #585: FormattedString(openTypeFeatures) not applied consistently
If I have a text that lays out on three pages or less, then the OT Features (ex. {liga}) get applied throughout. If the text is longer than three pages, then only the last three pages have features applied.

Example PDFs attached, using Source Serif 4 and {liga} feature to demonstrate. Short = three pages, Medium = five pages (pp. 1 & 2 don't get ligatures applied); Long = six pages (pp. 1, 2, & 3 don't have ligatures).

Script provided, along with sample text file that gets imported. (Path will need to get reconfigured.)

[SourceSerif4_testingOTFeatures-short.pdf](https://github.com/user-attachments/files/20266523/SourceSerif4_testingOTFeatures-short.pdf)

[SourceSerif4_testingOTFeatures-medium.pdf](https://github.com/user-attachments/files/20266524/SourceSerif4_testingOTFeatures-medium.pdf)

[SourceSerif4_testingOTFeatures-long.pdf](https://github.com/user-attachments/files/20266525/SourceSerif4_testingOTFeatures-long.pdf)

[spaceKernProof-testingOTFeatures.zip](https://github.com/user-attachments/files/20266527/spaceKernProof-testingOTFeatures.zip)

justvanrossum:
This fixes #585.


justvanrossum:
Working on a test case.

typemytype:
Will add a macOS version check as `kCTTypesetterOptionAllowUnboundedLayout` is only macOS 10.14+

I assume this is the main reason why it never came up...

justvanrossum:
Do we even still support macOS < 10.14?

typemytype:
I guess DrawBot still runs from 10.12+ ;)  (untested...) 

see https://github.com/typemytype/drawbot/pull/591

PR #598, opened 2026-02-21
[WIP] Add support to save apng

typemytype:


typemytype:
some testing needs to be added

typemytype:
argl tests are failing...

justvanrossum:
Are you sure you can't leverage Pillow's apng support more instead of implementing all this low level stuff?

https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#apng-sequences

typemytype:
DrawBot does not embed PIL, we only use it for testing.  

(it worked here locally)

typemytype:
reading all the errors: 

With .gif we support: 
* single frame ->  static .gif
* multiple frame -> animated .gif

if DrawBot does the same for png/apng, lots of scripts will generate animated pngs instead of the png of the last frame only. That is what the tests traceback indicate. 

possible solutions:
* fix all test by adding a `pageNumber` argument when a static png is required
* add a additional `animate` flag in `saveImage("myPath.png", animate=True)`, different from the .gif behavior (so .gif should change)

justvanrossum:
(Need to think about the api consequences)

There’s no reason we couldn’t bundle Pillow, no?

PR #599, opened 2026-05-19
adding align while calculating text bounds

typemytype:
`align` is an unused argument to calculate the text bounds but required if a plain string is provided instead of a FormatteddString


typemytype:
the failed booleanoperation tests are fixed in #583
not fixing this in this PR

justvanrossum:
Can you edit the description so it says what this PR fixes and/or why the change is needed?

typemytype:
edited

justvanrossum (approved):

